PR #2, opened 2019-04-30
reorganizes crate/adds brotli stream compression support

fairingrey:
## Description

This reorganizes the crate and adds brotli stream compression support.

## How Has This Been Tested?

Tests are added.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


src/brotli.rs
```diff
@@ -0,0 +1,87 @@
...
+        input_buffer.clear();
+
+        Poll::Ready(Some(Ok(compressed_output.freeze())))
+    }
+}
+
+pub fn compress_stream(
+    stream: impl Stream<Item = Result<Bytes>>,
+    compress: Compress,
+) -> impl Stream<Item = Result<Bytes>> {
```
Nemo157: If the struct is public you may as well name it here (alternatively, instead of a free function this could just be `CompressedStream::new`). There are some other traits it should expose if possible like `FusedStream`.

fairingrey: Good point, I'll change it in both mods then.

src/flate2.rs
```diff
@@ -0,0 +1,141 @@
+use core::{
```
Nemo157: I think this `flate2` module should be kept private and used by public `deflate`, `zlib` and `gzip` modules. That probably also means taking in just a `Compression` as the public config to set whether or not to use the header correctly.

Nemo157: (Obviously just adding a `deflate` module for now).

Nemo157: This can wait till later.

PR #7, opened 2019-05-01
Add proptest tests and test utilities module

Nemo157:
## Description
Adds automatically generated tests based on [`proptest`](https://altsysrq.github.io/proptest-book/intro.html) to ensure arbitrary compression works. Also refactors a lot of the existing tests to use common utilities to simplify them.

I'll probably move `PendRead` and `PendStream` upstream to `futures-test` since they're likely useful in other testing, so they can be pulled directly from there on the next release. (https://github.com/rust-lang-nursery/futures-rs/pull/1576)

## Motivation and Context

More thorough testing.

## Types of changes

## Checklist:

- [ ] All new and existing tests passed.

### DIFF
--- a/tests/utils/mod.rs
+++ b/tests/utils/mod.rs
@@ -1,0 +1,268 @@
+#![allow(unused)] // Different tests use a different subset of functions
+
+use bytes::Bytes;
+use futures::{
+    executor::{block_on, block_on_stream},
+    io::{AsyncBufRead, AsyncRead, AsyncReadExt},
+    stream::{self, Stream},
+};
+use pin_project::unsafe_project;
+use pin_utils::pin_mut;
+use proptest_derive::Arbitrary;
+use std::{
+    io::{self, Cursor, Read},
+    pin::Pin,
+    task::{Context, Poll},
+    vec,
+};
+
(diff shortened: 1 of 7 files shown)


tests/utils/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+impl<R: AsyncRead> AsyncRead for PendRead<R> {
+    fn poll_read(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut [u8],
+    ) -> Poll<io::Result<usize>> {
+        let this = self.project();
```
pickfire: I think we should get a variable that is self explanatory compared for `this`.

Nemo157: This should really be `let self = self.project();`, but you can't use `self` as a variable name, so `this` is just a synonym for it.

Nemo157:
Now that CI is running and decompression is all tested this should be good to merge, r? @fairingrey

fairingrey:
Looks good to me. I like how much cleaner the tests look to write. Feel free to merge it whenever @Nemo157

PR #10, opened 2019-05-02
Decompressed Stream Support

fairingrey:
## Description

This PR seeks to add support for decompressing streams of encoded bytes under each of the encoding algorithms used.

This is the current progress:

- [x] Brotli
- [x] Deflate
- [x] Gzip
- [x] Zlib

## Motivation and Context

Coming up with a standard pattern of decompressing encoded streams that the library already provides would be valuable for various use cases.

It would also unblock progress on https://github.com/rustasync/tide/pull/194.

## How Has This Been Tested?

Tests are added.

- [x] Brotli
- [x] Deflate
- [x] Gzip
- [x] Zlib

## Types of changes

## Checklist:

### DIFF
--- a/src/stream/gzip.rs
+++ b/src/stream/gzip.rs
@@ -170,9 +170,10 @@ fn get_header(level: Compression) -> Bytes {
 
 #[derive(Debug)]
 enum DeState {
+    ReadingHeader,
     Reading,
     Writing,
-    CheckingFooter,
+    ReadingFooter,
     Invalid,
 }
 
@@ -181,7 +182,6 @@ pub struct DecompressedGzipStream<S: Stream<Item = Result<Bytes>>> {
     #[pin]
     inner: S,
     state: DeState,
-    header_read: bool,
     input: BytesMut,
     output: BytesMut,
     crc: Crc,
@@ -219,11 +219,11 @@ impl<S: Stream<Item = Result<Bytes>>> Stream for DecompressedGzipStream<S> {
         #[allow(clippy::never_loop)] // https://github.com/rust-lang/rust-clippy/issues/4058
         loop {
             break match dbg!(mem::replace(this.state, DeState::Invalid)) {
-                DeState::Reading => {
+                DeState::ReadingHeader => {
                     *this.state = match ready!(this.inner.as_mut().poll_next(cx)) {
                         Some(chunk) => {
                             this.input.extend_from_slice(&chunk?);
-                            if !*this.header_read && this.input.len() >= 10 {
+                            if this.input.len() >= 10 {
                                 let header = this.input.split_to(10);
                                 if header[0..3] != [0x1f, 0x8b, 0x08] {
                                     return Poll::Ready(Some(Err(Error::new(
@@ -231,23 +231,27 @@ impl<S: Stream<Item = Result<Bytes>>> Stream for DecompressedGzipStream<S> {
                                         "Invalid file header",
                                     ))));
                                 }
-                                *this.header_read = true;
-                            }
-                            if !*this.header_read && this.input.len() < 10 {
                                 DeState::Reading
                             } else {
-                                DeState::Writing
+                                DeState::ReadingHeader
                             }
                         }
                         None => {
-                            if !*this.header_read {
-                                return Poll::Ready(Some(Err(Error::new(
-                                    ErrorKind::InvalidData,
-                                    "A valid header was not found",
-                                ))));
-                            }
-                            DeState::CheckingFooter
+                            return Poll::Ready(Some(Err(Error::new(
+                                ErrorKind::InvalidData,
+                                "A valid header was not found",
+                            ))));
                         }
+                    };
+                    continue;
+                }
+                DeState::Reading => {
+                    *this.state = match ready!(this.inner.as_mut().poll_next(cx)) {
+                        Some(chunk) => {
+                            this.input.extend_from_slice(&chunk?);
+                            DeState::Writing
+                        }
+                        None => DeState::ReadingFooter,
                     };
                     continue;
                 }
@@ -274,7 +278,7 @@ impl<S: Stream<Item = Result<Bytes>>> Stream for DecompressedGzipStream<S> {
                     Poll::Ready(Some(Ok(chunk)))
                 }
 
-                DeState::CheckingFooter => {
+                DeState::ReadingFooter => {
                     if this.input.len() == 8 {
                         let crc = &this.crc.sum().to_le_bytes()[..];
                         let bytes_read = &this.crc.amount().to_le_bytes()[..];
@@ -305,8 +309,7 @@ impl<S: Stream<Item = Result<Bytes>>> DecompressedGzipStream<S> {
     pub fn new(stream: S) -> DecompressedGzipStream<S> {
         DecompressedGzipStream {
             inner: stream,
-            state: DeState::Reading,
-            header_read: false,
+            state: DeState::ReadingHeader,
             input: BytesMut::new(),
             output: BytesMut::new(),
             crc: Crc::new(),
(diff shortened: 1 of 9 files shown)


src/stream/gzip.rs
```diff
@@ -42,15 +42,15 @@ impl<S: Stream<Item = Result<Bytes>>> Stream for GzipStream<S> {
...
+                    let bytes_read = &this.crc.amount().to_le_bytes()[..];
+                    footer.extend_from_slice(bytes_read);
                     *this.footer_appended = true;
                     return Poll::Ready(Some(Ok(footer)));
                 } else {
                     return Poll::Ready(None);
                 }
             } else if let Some(bytes) = ready!(this.inner.poll_next(cx)) {
-                *this.input_buffer = bytes?;
+                this.input_buffer.extend_from_slice(&bytes?);
```
Nemo157: Using `extend_from_slice` here will cause an unnecessary copy.

src/stream/gzip.rs
```diff
@@ -111,3 +111,96 @@ fn get_header(level: Compression) -> Bytes {
...
+                return Poll::Ready(None);
+            } else if let Some(bytes) = ready!(this.inner.poll_next(cx)) {
+                this.input_buffer.extend_from_slice(&bytes?);
+            } else {
+                *this.flushing = true;
+            }
+        }
+
+        if !*this.header_stripped {
+            let header = this.input_buffer.split_to(10);
```
Nemo157: This will panic if the first chunk contains less than 10 bytes.

Nemo157:
Overall looks good, just a couple of small issues in the flate compression.

src/stream/gzip.rs
fairingrey: Hmm, yeah... I was in the mindset that I would always be keeping the last 8 bytes so it would make it a necessity to do that but I might not have to.

src/stream/gzip.rs
fairingrey: Yup, though using the state machine implementation I think I might not have to deal with this problem (as if the chunk isn't large enough to strip out the header I can just keep reading more in)

fairingrey:
@Nemo157 I think I'm ready for another review

src/stream/gzip.rs
```diff
@@ -167,3 +167,150 @@ fn get_header(level: Compression) -> Bytes {
...
+        }
+
+        #[allow(clippy::never_loop)] // https://github.com/rust-lang/rust-clippy/issues/4058
+        loop {
+            break match dbg!(mem::replace(this.state, DeState::Invalid)) {
+                DeState::Reading => {
+                    *this.state = match ready!(this.inner.as_mut().poll_next(cx)) {
+                        Some(chunk) => {
+                            this.input.extend_from_slice(&chunk?);
+                            if !*this.header_read && this.input.len() >= 10 {
```
Nemo157: If you split this off to an initial `DeState::ReadingHeader` state then you can restrict the `extend_from_slice` to happening there and avoid the extra copying while reading the rest of the data.

tests/gzip.rs
```diff
@@ -48,3 +48,51 @@ fn gzip_stream_large() {
...
+    let data: Vec<u8> = data.unwrap().into_iter().flatten().collect();
+    assert_eq!(data, vec![1, 2, 3, 4, 5, 6]);
+}
+
+#[test]
+fn decompressed_gzip_stream_2() {
+    use async_compression::stream::gzip;
+
+    let stream = stream::iter(vec![
+        Bytes::from_static(&[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]),
```
Nemo157: For testing the behaviour here I think you'd want to pre-calculate the gzipped version of some data and then slice it in various ways to check that they're all dealt with correctly, e.g. the main ones I can think of:

```rust
let header: [u8; 10] = [...];
let body: [u8; _] = [...];
let trailer: [u8; 8] = [...];

// The entirety in one chunk
let stream = stream::iter(vec![
    Bytes::from(Vec::from_iter(header.iter().chain(body.iter()).chain(trailer.iter()).cloned())),
]);

// Header, body and trailer in separate chunks, similar to how `GzipStream` outputs it.
let stream = stream::iter(vec![
    Bytes::from(&header[..]),
    Bytes::from(&body[..]),
    Bytes::from(&trailer[..]),
]);

// Header, body and trailer each split across multiple chunks, no mixing
let stream = stream::iter(vec![
    Bytes::from(&header[0..5]),
    Bytes::from(&header[5..10]),
    Bytes::from(&body[0..half]),
    Bytes::from(&body[half..end]),
    Bytes::from(&trailer[0..4]),
    Bytes::from(&trailer[4..8]),
]);

// Header, body and trailer split across multiple chunks and mixed together
let stream = stream::iter(vec![
    Bytes::from(&header[0..5]),
    Bytes::from(Vec::from_iter(header[5..10].iter().chain(body[0..half].iter()).cloned())),
    Bytes::from(Vec::from_iter(body[half..end].iter().chain(trailer[0..4].iter()).cloned())),
    Bytes::from(&trailer[4..8]),
]);

File: src/stream/gzip.rs (hunk 3 above)
Comment (fairingrey) [reply src/stream/gzip.rs]: Oh, true, I'll try that.

File: tests/gzip.rs (hunk 4 above)
Comment (fairingrey) [reply tests/gzip.rs]: Oh, I can try this. Good suggestion!

**Comment by fairingrey** [comment]
A remark on using `Bytes::extend_from_slice` at https://github.com/rustasync/async-compression-rs/blob/014a2e101d9b044517399d9006872d84efb6c621/src/stream/gzip.rs#L255 -- the only time(s) this is reached is when the input is too small be read as data (as the footer will always take up 8 bytes, and the check on https://github.com/rustasync/async-compression-rs/blob/014a2e101d9b044517399d9006872d84efb6c621/src/stream/gzip.rs#L265 prevents the footer from being read as data). So I think this is the best I can get it without it freaking out on any odd/malformed data.

Let's go with one more review, and I think this should be good? @Nemo157

File: src/stream/gzip.rs [hunk 5]
```diff
@@ -167,3 +167,167 @@ fn get_header(level: Compression) -> Bytes {
...
+            crc.update(&output[0..output_len as usize]);
+            input.advance(input_len as usize);
+            Ok((status, output.split_to(output_len as usize).freeze()))
+        }
+
+        #[allow(clippy::never_loop)] // https://github.com/rust-lang/rust-clippy/issues/4058
+        loop {
+            break match mem::replace(this.state, DeState::Invalid) {
+                DeState::ReadingHeader => {
+                    *this.state = match ready!(this.inner.as_mut().poll_next(cx)) {
```
Nemo157: `ready!` hides an early `return Poll::Pending`, if the inner stream is not ready then this will return and leave the current state as `DeState::Invalid`. You should have a `*this.state = DeState::ReadingHeader;` before the match so that if it hits that early return it will go back to the current state (I was pretty undecided in #11 whether it was better to use `ready!` or to explicitly handle the `Poll` in match since this is kinda subtle).

src/stream/gzip.rs
```diff
@@ -167,3 +167,167 @@ fn get_header(level: Compression) -> Bytes {
...
+                            return Poll::Ready(Some(Err(Error::new(
+                                ErrorKind::InvalidData,
+                                "A valid header was not found",
+                            ))));
+                        }
+                    };
+                    continue;
+                }
+                DeState::Reading => {
+                    *this.state = match ready!(this.inner.as_mut().poll_next(cx)) {
```
Nemo157: Same issue with the early return here.

fairingrey:
Oh right, I forgot the inner stream can be pending. I'll fix it up right now.

Nemo157:
Awesome, now I'll rebase the proptest branch on this and see if I can break it 😈

PR #11, opened 2019-05-03
Fix compressing large chunks

Nemo157:
## Description
Flushing previously didn't guarantee that it would flush out the entire
data.

As part of this the flate and gzip implementations were refactored to a
state machine as I believe this makes them much more understandable,
especially the gzip implementation.

## Motivation and Context
This should fix the issue mentioned at https://github.com/rustasync/tide/pull/194#issuecomment-488900823

## How Has This Been Tested?
New tests added.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,3 +11,6 @@ bytes = "0.4.12"
 flate2 = "1.0.7"
 futures-preview = "0.3.0-alpha.15"
 pin-project = "0.3.2"
+
+[dev-dependencies]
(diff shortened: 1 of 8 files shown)


fairingrey:
Looks good. I think it makes a lot more sense now reading through the code. I will have to refactor https://github.com/rustasync/async-compression-rs/pull/10 but it shouldn't take too long, I imagine.

PR #13, opened 2019-05-11
Zstd support over Streams

fairingrey:
## Description

Adds support for compressing/decompressing Zstd over std futures' streams.

## Motivation and Context

Helps contribute to https://github.com/rustasync/async-compression-rs/issues/12. While Zstd isn't exactly well-known in the HTTP compression space, adding support for it would be nice for anyone who wants to use it, as it is comparable to (and at times, faster than) Brotli.

## How Has This Been Tested?

Tests are provided!

## Types of changes

## Checklist:

### DIFF
--- a/src/stream/zstd.rs
+++ b/src/stream/zstd.rs
@@ -1,0 +1,200 @@
+use std::{
+    io::Result,
+    mem,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+use bytes::{Bytes, BytesMut};
+use futures::{ready, stream::Stream};
+use pin_project::unsafe_project;
+use zstd::{
+    stream::raw::{Decoder, Encoder, Operation},
+    DEFAULT_COMPRESSION_LEVEL,
+};
+
+#[derive(Debug)]
+enum State {
+    Reading,
+    Writing(Bytes),
+    Flushing,
+    Done,
+    Invalid,
+}
+
+#[derive(Debug)]
+enum DeState {
+    Reading,
+    Writing(Bytes),
+    Done,
+    Invalid,
+}
+
+/// A zstd encoder, or compressor.
+///
+/// This structure implements a [`Stream`] interface and will read uncompressed data from an
+/// underlying stream and emit a stream of compressed data.
+#[unsafe_project(Unpin)]
+pub struct ZstdEncoder<S: Stream<Item = Result<Bytes>>> {
+    #[pin]
+    inner: S,
+    state: State,
+    output: BytesMut,
+    encoder: Encoder,
+}
+
+/// A zstd decoder, or decompressor.
+///
+/// This structure implements a [`Stream`] interface and will read compressed data from an
+/// underlying stream and emit a stream of uncompressed data.
+#[unsafe_project(Unpin)]
+pub struct ZstdDecoder<S: Stream<Item = Result<Bytes>>> {
+    #[pin]
+    inner: S,
+    state: DeState,
+    output: BytesMut,
+    decoder: Decoder,
+}
+
+impl<S: Stream<Item = Result<Bytes>>> ZstdEncoder<S> {
+    /// Creates a new encoder which will read uncompressed data from the given stream and emit a
+    /// compressed stream.
+    pub fn new(stream: S) -> ZstdEncoder<S> {
+        ZstdEncoder {
+            inner: stream,
+            state: State::Reading,
+            output: BytesMut::new(),
+            encoder: Encoder::new(DEFAULT_COMPRESSION_LEVEL).unwrap(),
+        }
+    }
+}
+
+impl<S: Stream<Item = Result<Bytes>>> ZstdDecoder<S> {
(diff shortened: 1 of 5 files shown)


fairingrey:
r? @Nemo157 

You can do it after https://github.com/rustasync/async-compression-rs/pull/14 is merged, since I rebased all my commits here on top of yours :smile:

fairingrey:
Actually, I still might probably change the API by including which encoding level to pick, similar to Brotli, but other than that I think it should be good to go.

src/stream/zstd.rs
```diff
@@ -0,0 +1,198 @@
...
+                    let chunk = compress(&mut this.encoder, &mut input, &mut this.output)?;
+
+                    *this.state = State::Writing(input);
+
+                    Poll::Ready(Some(Ok(chunk)))
+                }
+                State::Flushing => {
+                    let mut output = zstd_safe::OutBuffer::around(this.output);
+
+                    let bytes_left = this.encoder.flush(&mut output).unwrap();
```
Nemo157: Can `finish` be called here instead of having to first `flush` then `finish`? The Rust libs docs don't really seem to say, but the C docs have

> Calling ZSTD_compressStream2() with ZSTD_e_end instructs to finish a frame. It will perform a flush and write frame epilogue.

which is what I assume the Rust bindings will be calling into.

You might also need to resize the `output` in this state, if there's over 8k of data to flush for some reason then you'll consume the entire output buffer on the first poll and have no space available the next time.

Nemo157:
LGTM other than that one comment  👍

src/stream/zstd.rs
fairingrey: So interesting story here, but I actually thought it would do the same -- I was repeatedly calling finish until I realized it didn't have the right behavior, because as I kept calling it, the bytes left in the internal buffer remained the same (32009).

Curiously enough, `flush` has the right behavior, but doesn't finish the final internal frame (as the Zstd decoder reaches unexpected EOF and returns the proper error without [line 134](https://github.com/rustasync/async-compression-rs/pull/13/files#diff-c3c3f8cc95ca9e4ee4eb756858c8a398R134)), hence the final call to finish. Actually, the boolean included doesn't even really matter in the code, it seems:

```rust
    fn finish(
        &mut self,
        output: &mut OutBuffer,
        _finished_frame: bool,
    ) -> io::Result<usize> {
        zstd_safe::end_stream(&mut self.context, output).map_err(map_error_code)
    }
```

Understandably it does look a bit strange, but when you debug the code as follows:

```rust
State::Flushing => {
    dbg!(&this.output.len()); // note the dbg here
    let mut output = zstd_safe::OutBuffer::around(this.output);

    let bytes_left = this.encoder.flush(&mut output).unwrap();
    dbg!(&output.pos, &bytes_left); // and here
    *this.state = if bytes_left == 0 {
        let _ = this.encoder.finish(&mut output, true);
        State::Done
    } else {
        State::Flushing
    };
    Poll::Ready(Some(Ok(output.as_slice().into())))
}
```
This is what the tests spit out with `cargo test zstd_stream_compress -- --nocapture`

```
running 2 tests
[src/stream/zstd.rs:130] &this.output.len() = 8000
[src/stream/zstd.rs:134] &output.pos = 15
[src/stream/zstd.rs:134] &bytes_left = 0
test zstd_stream_compress ... ok
[src/stream/zstd.rs:130] &this.output.len() = 8000
[src/stream/zstd.rs:134] &output.pos = 8000
[src/stream/zstd.rs:134] &bytes_left = 32009
[src/stream/zstd.rs:130] &this.output.len() = 8000
[src/stream/zstd.rs:134] &output.pos = 8000
[src/stream/zstd.rs:134] &bytes_left = 24009
[src/stream/zstd.rs:130] &this.output.len() = 8000
[src/stream/zstd.rs:134] &output.pos = 8000
[src/stream/zstd.rs:134] &bytes_left = 16009
[src/stream/zstd.rs:130] &this.output.len() = 8000
[src/stream/zstd.rs:134] &output.pos = 8000
[src/stream/zstd.rs:134] &bytes_left = 8009
[src/stream/zstd.rs:130] &this.output.len() = 8000
[src/stream/zstd.rs:134] &output.pos = 8000
[src/stream/zstd.rs:134] &bytes_left = 9
[src/stream/zstd.rs:130] &this.output.len() = 8000
[src/stream/zstd.rs:134] &output.pos = 9
[src/stream/zstd.rs:134] &bytes_left = 0
test zstd_stream_compress_large ... ok
```
You can note here that the "OutBuffer" that wraps around `this.output` actually just uses it as a raw byte buffer that repeatedly fills it to maximum capacity (8000) on every `flush` call when the buffer can't consume all of the bytes left in the internal buffer. Nothing actually moves in `this.output` since `output` (the outbuffer) keeps track of its' own internal position within the buffer, so it doesn't need to be resized.

Hopefully that explains my reasoning well enough, but yeah -- it was quite an interesting ride to figure out.

PR #15, opened 2019-05-12
Add access to the underlying readers/streams for all encoders/decoders

Nemo157:
## Motivation and Context
When the readers/streams are known to be some other concrete type in usage it's relatively common to want to call other methods on them that won't affect the compression (e.g. the concrete type may have a method that tells you the current position and total length so you can see how far through the stream you are).

## How Has This Been Tested?

The type system is strong enough to ensure these methods are correct.

## Types of changes

## Checklist:

### DIFF
--- a/src/bufread/deflate.rs
+++ b/src/bufread/deflate.rs
@@ -33,6 +33,35 @@ impl<R: AsyncBufRead> DeflateEncoder<R> {
             compress: Compress::new(level, false),
         }
     }
+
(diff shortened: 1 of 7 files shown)


Nemo157:
I'll add `zstd` support in a later PR now that that's in master.

r? @fairingrey

fairingrey:
Skimmed over it, looks good! And yeah, you could do that.

PR #17, opened 2019-05-13
Allow using !Unpin streams with zlib/deflate

Nemo157:
## Description
Remove `Unpin` bounds from `ZlibEncoder`/`DeflateEncoder`.

## Motivation and Context
More general for potential usecases.

## Types of changes

## Checklist:

### DIFF
--- a/src/stream/flate.rs
+++ b/src/stream/flate.rs
@@ -20,7 +20,7 @@ enum State {
 }
 
 #[unsafe_project(Unpin)]
-pub(crate) struct FlateEncoder<S: Stream<Item = Result<Bytes>>> {
+pub struct FlateEncoder<S: Stream<Item = Result<Bytes>>> {
     #[pin]
     inner: S,
     state: State,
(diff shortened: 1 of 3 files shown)


src/stream/flate.rs
```diff
@@ -20,7 +20,7 @@ enum State {
 }
 
 #[unsafe_project(Unpin)]
-pub(crate) struct FlateEncoder<S: Stream<Item = Result<Bytes>>> {
+pub struct FlateEncoder<S: Stream<Item = Result<Bytes>>> {
```
Nemo157: It's unfortunate that this needs to be `pub` for the `Unpin` bounds that are added. It's not actually public because the `flate` module is private, but I'll likely open an upstream issue about the implementation detail leak that this forces.

fairingrey: Yeah, I was wondering a little bit about that. Since it's not actually exposed through an existing module through `pub use` or `pub mod`, it won't be shown, but I do get the feeling it might seem a bit strange.

It's fine for now though, I think.

PR #20, opened 2019-05-13
Add features to cover different usecases

Nemo157:
## Description
See the updated docs in lib.rs for details on the design.

## Motivation and Context
This allows users to only pay compilation cost for the compressions and usecases they care about (e.g. Tide can now stop compiling the `bufread` implementations it doesn't use).

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,14 +11,24 @@ description = """
 Adaptors between compression crates and Rust's modern asynchronous IO types.
 """
 
+[features]
+default = ["bufread", "stream", "brotli", "deflate", "gzip", "zlib", "zstd"]
+bufread = []
+stream = ["bytes"]
+brotli = ["brotli2"]
+deflate = ["flate2"]
+gzip = ["flate2"]
+zlib = ["flate2"]
+zstd = ["libzstd", "zstd-safe"]
+
 [dependencies]
-brotli2 = "0.3.2"
-bytes = "0.4.12"
-flate2 = "1.0.7"
+brotli2 = { version = "0.3.2", optional = true }
+bytes = { version = "0.4.12", optional = true }
+flate2 = { version = "1.0.7", optional = true }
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -11,14 +11,24 @@ description = """
...
-bytes = "0.4.12"
-flate2 = "1.0.7"
+brotli2 = { version = "0.3.2", optional = true }
+bytes = { version = "0.4.12", optional = true }
+flate2 = { version = "1.0.7", optional = true }
 futures-preview = "0.3.0-alpha.16"
 pin-project = "0.3.2"
-zstd = "0.4"
-zstd-safe = "1.4"
+libzstd = { version = "0.4", optional = true, package = "zstd", default-features = false }
```
Nemo157: The renaming of the package here is a little bit of a hack to allow a feature with the same name as an optional dependency.

fairingrey: Fair enough, I don't mind the disambiguation here.

fairingrey:
@Nemo157 Travis freaks out over a `cargo fmt --all -- --check`, want to just run that real quick? I can also commit the change to this PR, I believe, but just checking with you.

src/stream/zstd.rs
```diff
@@ -9,7 +9,7 @@ use std::{
 use bytes::{Bytes, BytesMut};
 use futures::{ready, stream::Stream};
 use pin_project::unsafe_project;
```
fairingrey: Travis freaks out on this line, `use pin_project::unsafe_project;` should be below libzstd

fairingrey:
Other than that, it's nice they're all included by default but can be reduced to minimal dependencies if a user calls for it. I'm a fan of the change.

Nemo157:
@fairingrey thanks for fixing that up! I should probably have run the tests before actually checking that box...

PR #22, opened 2019-05-18
Add AsyncWrite variants

Nemo157:
## Description
Adds implementation of `AsyncWrite` wrapping encoder and decoders to round out the likely usecases.

(Has been rebased and based on abstraction from #34, so blocked on that merging first).

## Motivation and Context
For some usecases you may have an `AsyncWrite` that you want to automatically compress/decompress as you write to.

## How Has This Been Tested?
Tests added.

## Types of changes

## Checklist:


src/write/buf_write.rs
```diff
@@ -0,0 +1,29 @@
+use std::{
+    io,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+pub(crate) trait AsyncBufWrite {
+    /// Attempt to return an internal buffer to write to, flushing data out to the inner reader if
```
danielhenrymantilla: Given the implementation that follows, I'd slightly reword this as follows:
```suggestion
    /// Attempt to return an internal buffer to write to, flushing pending `produce`d
    /// data out to the inner writer.
```

src/write/buf_write.rs
```diff
@@ -0,0 +1,29 @@
...
+    /// it is full.
```
danielhenrymantilla:
```suggestion
```

src/write/buf_write.rs
```diff
@@ -0,0 +1,29 @@
...
+    ///
+    /// On success, returns `Poll::Ready(Ok(buf))`.
+    ///
+    /// If the buffer is full and cannot be flushed, the method returns `Poll::Pending` and
```
danielhenrymantilla:
```suggestion
    /// If the buffer's `pending` data cannot be flushed, the method returns `Poll::Pending` and
```

src/write/buf_write.rs
```diff
@@ -0,0 +1,29 @@
...
+    /// arranges for the current task context (`cx`) to receive a notification when the object
+    /// becomes readable or is closed.
```
danielhenrymantilla:
```suggestion
    /// becomes writeable or is closed.
```

src/write/buf_write.rs
```diff
@@ -0,0 +1,29 @@
...
+    fn poll_flush_buf(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<&mut [u8]>>;
+
+    /// Tells this buffer that `amt` bytes have been written to its buffer, so they should be
+    /// written out to the underlying IO when possible.
```
danielhenrymantilla:
```suggestion
    /// written out to the underlying writer when flushed.
```

src/write/buf_writer.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// Gets a mutable reference to the underlying writer.
+    ///
+    /// It is inadvisable to directly write to the underlying writer.
+    pub fn get_mut(&mut self) -> &mut W {
+        &mut self.inner
+    }
+
+    /// Gets a pinned mutable reference to the underlying writer.
+    ///
+    /// It is inadvisable to directly write to the underlying writer.
```
danielhenrymantilla:
```suggestion
    /// It is inadvisable to directly write to the underlying writer unless flushed
    /// beforehand.
```

src/write/buf_writer.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub fn get_pin_mut(self: Pin<&mut Self>) -> Pin<&mut W> {
+        self.project().inner
+    }
+
+    /// Consumes this `BufWriter`, returning the underlying writer.
+    ///
+    /// Note that any leftover data in the internal buffer is lost.
```
danielhenrymantilla:
```suggestion
    /// Note that any leftover data in the internal buffer is lost unless flushed
    /// beforehand.
```

src/write/buf_write.rs
Nemo157: This could be a direct `AsyncBufWrite` implementation on some IO device, it might not be a wrapper around another writer (I shouldn't have mentioned an "inner reader" earlier either).

src/write/buf_write.rs
Nemo157: heh, this mistake has been copy-pasted across so much documentation by now.

src/write/buf_write.rs
Nemo157: This only requires a partial flush, I'll have a think about how to make the semantics a bit clearer.

PR #28, opened 2019-08-16
Make ZstdEncoder/Decoder Sync

### LINKED ISSUE #27: ZstdDecoder is !Sync
`ZstdDecoder` is [currently `!Sync`](https://docs.rs/async-compression/0.1.0-alpha.1/async_compression/stream/struct.ZstdDecoder.html#impl-Sync) because [`zstd::stream::raw::Decoder` is `!Sync`](https://docs.rs/zstd/0.4.27+zstd.1.4.2/zstd/stream/raw/struct.Decoder.html#impl-Sync). This is limiting because some crates like `hyper` want `+ Send + Sync` for their bytes streams.

As far as I know, it should be sound to add `unsafe impl<S> Sync for ZstdDecoder<S> where S: Sync` because we don't access the decoder in any `&self` method (except maybe `Debug` which would have to be checked).

If that is sound, it would probably be safer to wrap the decoder in a type that only gives unique access like:

```rust
struct Unshared<T>(T);

impl<T> Unshared<T> {
    pub fn new(inner: T) -> Self { Unshared(inner) }

    pub fn get_mut(&mut self) -> &mut T { &mut self.0 }

    pub fn into_inner(self) -> T { self.0 }
}

unsafe impl<T> Sync for Unshared<T> {}
```

Nemo157:
## Motivation and Context

fixes #27 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,0 +1,2 @@
+pub fn _assert_send<T: Send>() {}
+pub fn _assert_sync<T: Sync>() {}
--- a/src/bufread/deflate.rs
+++ b/src/bufread/deflate.rs
@@ -96,3 +96,8 @@ impl<R: AsyncBufRead> AsyncRead for DeflateEncoder<R> {
         }
     }
 }
+
+fn _assert() {
+    crate::util::_assert_send::<DeflateEncoder<Pin<Box<dyn AsyncBufRead + Send>>>>();
+    crate::util::_assert_sync::<DeflateEncoder<Pin<Box<dyn AsyncBufRead + Sync>>>>();
+}
--- a/src/bufread/zlib.rs
+++ b/src/bufread/zlib.rs
@@ -96,3 +96,8 @@ impl<R: AsyncBufRead> AsyncRead for ZlibEncoder<R> {
         }
     }
 }
+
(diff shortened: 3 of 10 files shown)


src/util.rs
```diff
@@ -0,0 +1,26 @@
...
+        Unshared { inner }
+    }
+
+    pub fn get_mut(&mut self) -> &mut T {
+        &mut self.inner
+    }
+}
+
+// Safety: we do not access `self.inner` in any `&self` method, so there is no way to have shared
+// references to it in multiple threads.
```
Nemo157: This is incorrect with scoped threads, you could do something like

```rust
let mut foo = Unshared::new(...);
let bar = &*foo.get_mut();
spawn(move || { dbg!(bar) });
spawn(move || { dbg!(bar) });
```

Nemo157: Oh, actually no, I'm wrong. Once you go to `&T` that is `!Send` because `T: !Sync`. So I think this is fine.

Nemo157:
Some folks on the community discord #black-magic took a look and seem to think this is sound.

PR #45, opened 2019-10-06
[bzip2] Initial implementation

oleid:
This is an attempt to add bzip2 support to this library. It's at RFC stage, since a few unreleased  bzip2 changes are still needed, c.f. https://github.com/alexcrichton/bzip2-rs/pull/49. 

EDIT: It now works with the latest stable version

## Description
Since flate and bzip2 crates are quite similar, I used that as a basis for the implementation. Only details are different.

## How Has This Been Tested?
I added a test case. It still fails, but I'm not entirely sure why. Probably something is wrong at compression, since I can decompress just fine using the crate in a (still private) app. Maybe you have some more insight? 

EDIT: Tests pass fine :).

## Types of changes
- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have read the [CONTRIBUTING](https://github.com/rustasync/tide/blob/master/.github/CONTRIBUTING.md) document.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/src/codec/bzip2/decoder.rs
+++ b/src/codec/bzip2/decoder.rs
@@ -1,0 +1,80 @@
+use crate::{codec::Decode, util::PartialBuffer};
+use std::fmt;
+use std::io::{Error, ErrorKind, Result};
+
+use bzip2::{Decompress, Status};
+
+pub struct BzDecoder {
+    decompress: Decompress,
+}
+
+impl fmt::Debug for BzDecoder {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "BzDecoder {{total_in: {}, total_out: {}}}",
+            self.decompress.total_in(),
+            self.decompress.total_out()
+        )
+    }
+}
+
+impl BzDecoder {
+    pub(crate) fn new() -> Self {
+        Self {
+            decompress: Decompress::new(false),
+        }
+    }
+
+    fn decode(
+        &mut self,
+        input: &mut PartialBuffer<&[u8]>,
+        output: &mut PartialBuffer<&mut [u8]>,
+    ) -> Result<Status> {
+        let prior_in = self.decompress.total_in();
+        let prior_out = self.decompress.total_out();
+
+        let status = self
+            .decompress
+            .decompress(input.unwritten(), output.unwritten_mut())
+            .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
+
+        input.advance((self.decompress.total_in() - prior_in) as usize);
+        output.advance((self.decompress.total_out() - prior_out) as usize);
+
+        Ok(status)
+    }
+}
+
+impl Decode for BzDecoder {
+    fn decode(
+        &mut self,
+        input: &mut PartialBuffer<&[u8]>,
+        output: &mut PartialBuffer<&mut [u8]>,
+    ) -> Result<bool> {
+        match self.decode(input, output)? {
+            // Decompression went fine, nothing much to report.
+            Status::Ok => Ok(false),
+
+            // The Flush action on a compression went ok.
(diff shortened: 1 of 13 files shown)


Nemo157:
The tests are quite aggressive about using small input and output chunk sizes (only 2 bytes in a lot of cases) to try and trigger edge cases, so it's quite likely that it can work in a real usecase but fail the tests.

Looking at what happens it looks like bzip2 decompression doesn't distinguish between end of input and completion of flushing output, so making `BzDecoder::finish` just return `Ok(true)` passes all of the tests.

For compression I think `FinishOk` just means it successfully flushed some output data, but there is still data remaining to flush, once it's complete it returns `StreamEnd`. Swapping the return values for those statuses got all the compression tests passing.

You can see the debugging I added to find that, and the changes that made it pass here: <https://github.com/Nemo157/async-compression-rs/commit/f009b68ea02225bbf51dc6b7049979aad3de7532>

src/codec/bzip2/decoder.rs
```diff
@@ -0,0 +1,98 @@
...
+            // The stream's end has been met, meaning that no more data can be input.
+            Status::StreamEnd => Ok(true),
+
+            // There was insufficient memory in the input or output buffer to complete
+            // the request, but otherwise everything went normally.
+            Status::MemNeeded => Err(Error::new(ErrorKind::Other, "out of memory")),
+        }
+    }
+
+    fn finish(&mut self, output: &mut PartialBuffer<&mut [u8]>) -> Result<bool> {
```
Nemo157: So this should just return `Ok(true)` since it was already finished during the `decode`.

src/codec/bzip2/encoder.rs
```diff
@@ -0,0 +1,119 @@
...
+            Status::Ok => Ok(false),
+
+            // The Flush action on a compression went ok.
+            Status::FlushOk => unreachable!(),
+
+            // The Run action on compression went ok.
+            Status::RunOk => unreachable!(),
+
+            // The Finish action on compression went ok.
+            Status::FinishOk => Ok(true),
```
Nemo157: This should be `Ok(false)`

src/codec/bzip2/encoder.rs
```diff
@@ -0,0 +1,119 @@
...
+
+            // The stream's end has been met, meaning that no more data can be input.
+            Status::StreamEnd => unreachable!(),
```
Nemo157: And this should be `Ok(true)`

oleid:
Thanks for taking the time to debug the issue!

I guess we could manually convert the error for now; not sure why I haven't done that in the first place ;)

Anyway, would you like me to clean up the commits some more (I kept them separate for further review), or shall we just squash them together during merge?

Nemo157:
Thanks for implementing this!

Feel free to squash then `bors r+` when it's ready.

bors delegate+

Nemo157:
Oh, just remembered about proptest tests as well, it'd be great if you could add `bzip2` to the algos tested here as well

https://github.com/rustasync/async-compression/blob/c5d9bcb0d3a87aaaa034370cc7d1418c67bc52ee/tests/proptest.rs#L61

oleid:
> Oh, just remembered about proptest tests as well, it'd be great if you could add `bzip2` to the algos tested here as well
> 
> https://github.com/rustasync/async-compression/blob/c5d9bcb0d3a87aaaa034370cc7d1418c67bc52ee/tests/proptest.rs#L61

Done!

oleid:
bors +r

oleid:
> Did you mean "r+"?

Yes, I did!

bors "r+"

oleid:
Oh brave new world, that has such wonderful bots in it! :smile:

PR #48, opened 2019-10-14
Elide lifetimes in `Pin<&(mut) Self>`

taiki-e:


Nemo157:
Thanks, bors r+

Nemo157:
:( I guess bors is start of line only

bors r+

Issue #59, opened 2019-11-14
Rename `bzip` feature back to `bzip2`


Nemo157:
I should really read the history of formats before making arbitrary changes like that. Will have to keep the feature until `0.2`, but can deprecate it and have it just activate the correct `bzip2` feature, and rename it everywhere in the tests.

Issue #56, opened 2019-11-14
Move bufread/write under futures module


Nemo157:
Until the whole split ecosystem around `futures-io` and `tokio-io` traits is sorted out it'd make sense to support both. The current `bufread` and `write` modules can be moved under a `futures` module (and a `#[doc(hidden)] #[deprecated]` re-export added) to allow adding `tokio::{bufread, write}` modules once a non-alpha `tokio-io 0.2` release is available (and maybe some preview modules before that).

PR #61, opened 2019-12-05
Upgrade bytes to 0.5

sfackler:
This is a breaking change to the stream interface, but synchronizes the
version of Bytes to what the upcoming hyper release is using.


Nemo157:
LGTM, thanks!

I’m gonna quickly do a deprecation release for #56 and #59, then this can go into 0.2.

Nemo157:
Thanks again, I rebased and applied `cargo fmt` to fix the `use` ordering.

bors r+

PR #69, opened 2020-01-16
Remove trait bounds on type definitions

tesaguri:
This change allows downstream types to depend on `async_compression` types without extra trait bounds.

Currently, users need to write something like the following:

```rust
struct Foo<S: Stream<Item = Result<Bytes>>> {
    stream: GzipDecoder<S>,
    // ...
}
```

After this change, it becomes:

```rust
struct Foo<S> {
    stream: GzipDecoder<S>,
    // ...
}
```

The trait bounds are still enforced on the constructors, so they still prevent you from creating `*Decoder<T>` or `*Encoder<T>` values when the required trait bounds are not satisfied.

### DIFF
--- a/src/futures/bufread/generic/decoder.rs
+++ b/src/futures/bufread/generic/decoder.rs
@@ -18,7 +18,7 @@ enum State {
 
 pin_project! {
     #[derive(Debug)]
-    pub struct Decoder<R: AsyncBufRead, D: Decode> {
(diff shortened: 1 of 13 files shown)


Nemo157:
I was really hopeful that implied trait bounds would one day become a thing, but there's been very little movement on that, so this LGTM.

bors r+

PR #70, opened 2020-01-24
Replace `brotli2` with `brotli`.

### LINKED ISSUE #39: Brotli implementation choice
Currently the [`brotli2`](https://crates.io/crates/brotli2) crate is used for brotli compression, which has a `-sys` dependency. These crates can be inconvenient to deal with because of `links=`, and `brotli-sys` also bundles a relatively outdated version of [google/brotli](https://github.com/google/brotli). Would it be reasonable to support the [`brotli`](https://crates.io/crates/brotli) crate instead of or in addition to `brotli2`?

daxpedda:
Fixes #39.

I removed `brotli2` because it is severely outdated. It uses a version of googles `brotli` that is almost 3 years old.
If requested I can add `brotli2`, but would like to have some guidance on naming the features.

### DIFF
--- a/src/codec/brotli/decoder.rs
+++ b/src/codec/brotli/decoder.rs
@@ -4,16 +4,20 @@ use std::{
     io::{Error, ErrorKind, Result},
 };
 
-use brotli2::raw::{DeStatus, Decompress};
+use brotli::{enc::StandardAlloc, BrotliDecompressStream, BrotliResult, BrotliState};
 
 pub struct BrotliDecoder {
-    decompress: Decompress,
+    state: BrotliState<StandardAlloc, StandardAlloc, StandardAlloc>,
 }
 
 impl BrotliDecoder {
     pub(crate) fn new() -> Self {
         Self {
-            decompress: Decompress::new(),
+            state: BrotliState::new(
+                StandardAlloc::default(),
+                StandardAlloc::default(),
+                StandardAlloc::default(),
+            ),
         }
     }
 
@@ -21,17 +25,28 @@ impl BrotliDecoder {
         &mut self,
         input: &mut PartialBuffer<&[u8]>,
         output: &mut PartialBuffer<&mut [u8]>,
-    ) -> Result<DeStatus> {
-        let mut in_buf = input.unwritten();
+    ) -> Result<BrotliResult> {
+        let in_buf = input.unwritten();
         let mut out_buf = output.unwritten_mut();
 
-        let original_input_len = in_buf.len();
-        let original_output_len = out_buf.len();
+        let mut input_len = 0;
+        let mut output_len = 0;
 
-        let status = self.decompress.decompress(&mut in_buf, &mut out_buf)?;
-
-        let input_len = original_input_len - in_buf.len();
-        let output_len = original_output_len - out_buf.len();
+        let status = match BrotliDecompressStream(
+            &mut in_buf.len(),
+            &mut input_len,
+            &in_buf,
+            &mut out_buf.len(),
+            &mut output_len,
+            &mut out_buf,
+            &mut 0,
+            &mut self.state,
+        ) {
+            BrotliResult::ResultFailure => {
+                return Err(Error::new(ErrorKind::Other, "brotli error"))
+            }
+            status => status,
+        };
 
         input.advance(input_len);
         output.advance(output_len);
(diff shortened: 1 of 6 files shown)


src/codec/brotli/encoder.rs
```diff
@@ -1,38 +1,58 @@
...
+            &mut out_buf.len(),
+            &mut out_buf,
+            &mut output_len,
+            &mut None,
+            &mut |_, _, _, _| (),
+        ) <= 0
+        {
+            return Err(Error::new(ErrorKind::Other, "brotli error"));
+        } else {
+            self.state.available_out_ == 0
```
daxpedda: This is supposed to be `BrotliEncoderIsFinished`, but it doesn't work if there are multiple streams and never enters `BROTLI_STREAM_FINISHED` state.
This may be a bug in `brotli` or fixable in `async-compression`, I'm not sure.

dekellum:
> I removed brotli2 because it is severely outdated. 

Note that this makes this a breaking change. I guess if "severely" means you suspect this is a security liability that could be justified. If and when alexcrichton/brotli2-rs#26 happens, brotli2 could be resurrected here.

daxpedda:
Actually glancing over the commits I found only this:
-  C: enc: fix rare access to uninitialized data in ring-buffer (google/brotli@4b2b2d4f83ffeaac7708e44409fe34896a01a278)
-  platform: fix unaligned 64-bit accesses on AArch32 (google/brotli@d4cd6cdf1cd52b75f871b6d75b1c8941454371db)
-  fix "left shift before promotion" (google/brotli@68db5c0272d46f316efab575b67dae5b9ce4378d)
-  fix unaligned access for `aarch64`-cross-`armhf` build (google/brotli@0f3c84e7458d2ef91a29fdf269e8ad016ae694ba)

Only the first one seems likely to be a security issues, I don't know anything about brotli to judge the rest.

With "severely", which was a poor choice of words, I actually meant all the build issues it had, which I already encountered on a server of mine.

dekellum:
Thanks for looking, which is complimentary to my look at the brotli release notes. Just to be clear, I am quite pleased to see a `brotli` integration land here! I am only questioning the need to remove `brotli2`.  See my comments in #39.

Nemo157:
Personally I dislike abusing features to select between alternative implementations, it doesn’t really work with how features are designed. So unless there are major reasons to keep both (in which case I’d probably go for having both available under different feature and type names) I’d prefer to just have a single implementation.

At this point doing a breaking change to switch it is fine, I understood it would likely happen when exposing the configuration types, and should probably open an issue about coming up with a plan to hide implementation details like that.

I should have time tomorrow to review in detail.

dekellum:
Just to summarize my personal opinion/rationale for wanting both: 
* `brotli`: pure rust so likely more safe, easier to compile, kept up to date.
* `brotli2`: the original implementation, likely faster (suspect from blog posts on the former, would know based on benchmarks if all available.)

I don't really follow your concerns about features. Maybe that has something to do with how features are being used here, that I don't yet understand?

dekellum:
Also FWIW, I'm trying to do this: alexcrichton/brotli2-rs#31 . If I (also) fail, then that weakens my case, I guess.

daxpedda:
> At this point doing a breaking change to switch it is fine, I understood it would likely happen when exposing the configuration types, and should probably open an issue about coming up with a plan to hide implementation details like that.

I could easily implement a copy of `CompressParams` to replace the public `BrotliEncoderParams`. This would turn this into a non-break-change then.

Nemo157:
> I could easily implement a copy of `CompressParams` to replace the public `BrotliEncoderParams`. This would turn this into a non-break-change then.

Unfortunately that would still be a breaking change, it's publicly observable that `async-compression::brotli2::CompressParams == brotli2::CompressParams`. Something like that is probably the way to go to make future changes like this non-breaking, and probably should be done in the same release, but would be best as a separate PR.

Nemo157:
> I don't really follow your concerns about features. Maybe that has something to do with how features are being used here, that I don't yet understand?

Currently all the features here are intended to enable specific _algorithms_ and the implementations used are opaque (ignoring the current leak through the configuration types). So one idea to allow selecting the backing implementation via features would be something like

 * `brotli` enables the `BrotliEncoder` type
 * `brotli-impl-brotli` makes it use `brotli` as the implementation
 * `brotli-impl-brotli2` makes it use `brotli2` as the implementation

But then the question is, what happens when within one build crate `foo` enables features `["brotli", "brotli-impl-brotli"]` and crate `bar` enables features `["brotli", "brotli-impl-brotli2"]`? Do we have a precedence order between the two so that we select one to use, even though we'll still be compiling both as dependencies?

Cargo features just are not designed to work for choosing between things like this, they are intended to be purely additive and just enable new bits of API.

If `brotli2` is truly needed for whatever reason, I would want it to be done via a `brotli2` feature which enables a separate `Brotli2Encoder` type, then if `foo` uses the `brotli` feature and `bar` uses the `brotli2` feature, they would just end up using distinct parts of the API.

src/codec/brotli/decoder.rs
```diff
@@ -4,34 +4,49 @@ use std::{
...
-        let original_output_len = out_buf.len();
+        let mut input_len = 0;
+        let mut output_len = 0;
 
-        let status = self.decompress.decompress(&mut in_buf, &mut out_buf)?;
-
-        let input_len = original_input_len - in_buf.len();
-        let output_len = original_output_len - out_buf.len();
+        let status = match BrotliDecompressStream(
+            &mut in_buf.len(),
```
Nemo157: I assume the modifications to `available_in`/`available_out` just mirror the modifications to `input_offset`/`output_offset`, so we can ignore them?

src/codec/brotli/encoder.rs
Nemo157: This is because the `FINISHED` state is not hit when calling `flush`, changing this to

```rust
if op == BrotliEncoderOperation::BROTLI_OPERATION_FLUSH {
    BrotliEncoderHasMoreOutput(&self.state) == 0
} else {
    BrotliEncoderIsFinished(&self.state) != 0
}
```

works, though maybe it'd be better to return both those checks as flags that can be checked in `flush` and `finish`.

src/codec/brotli/decoder.rs
Nemo157: Yeah, after looking a bit more into how `brotli` works this looks like the case.

Nemo157:
LGTM with the finished state fix, I have an idea about how to maybe do a deprecation cycle with the parameters so that there can be a straight-forward upgrade. I'll try and look into that in the next day or two so I'll hold off merging this till I work out what's happening there to avoid juggling an old version branch.

src/codec/brotli/encoder.rs
daxpedda: I just moved the checks down to `flush` and `finish`.

daxpedda:
> Unfortunately that would still be a breaking change, it's publicly observable that `async-compression::brotli2::CompressParams == brotli2::CompressParams`.

I thought considering that we are re-exporting this, the fact that these two types are the same would be an implementation detail?
But I can imagine it still being a breaking change for some people out there.

Nemo157:
The re-exporting was mostly a convenience, so that users didn't need to depend on `brotli2` as well if they wanted to customise it. I doubt the change would break anyone, but that also means doing a breaking release isn't a big deal since most users should be able to update without any code changes.

Fitting in a deprecation cycle feels like it's more annoying than it's worth, so I'll merge this and should get to implementing #71 and doing a release this week.

@bors r+

PR #73, opened 2020-01-27
Remove implementation details to encoder configuartions.

### LINKED ISSUE #71: Hide implementation details
Currently the `flate2` and `brotli2` (soon to be `brotli`, see #70) libraries are exposed publicly through configuring the constructors of the `Encoder`s using them. This should be hidden so that changing which library is used is not a breaking change.

Main idea would be to replace the re-exports with locally defined configuration parameters, then map those to the underlying implementation internally.

daxpedda:
Fixes #71.

This turned out to have some issues:
-  this way a user can change the `quality` while encoding is in process, some proposals:
   -  if we want to keep the build pattern, maybe we should do it like other builder patterns, with a `build` function, so users can't change configuration after it's set
   -  just document it and expect users not to do this
-  should I do the same with all other implementations?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -141,13 +141,6 @@ pub mod flate2 {
     pub use flate2::Compression;
 }
 
-/// Types to configure [`brotli`](::brotli) based encoders.
-#[cfg(feature = "brotli")]
-#[cfg_attr(docsrs, doc(cfg(feature = "brotli")))]
-pub mod brotli {
-    pub use brotli::enc::backward_references::BrotliEncoderParams;
-}
-
 /// Types to configure [`bzip2`](::bzip2) based encoders.
 #[cfg(feature = "bzip2")]
 #[cfg_attr(docsrs, doc(cfg(feature = "bzip2")))]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -134,9 +134,12 @@ pub mod stream;
 mod unshared;
 mod util;
 
+#[cfg(feature = "brotli")]
+use brotli::enc::backward_references::BrotliEncoderParams;
+
 /// Level of compression data should be compressed with.
 #[derive(Clone, Copy, Debug)]
-pub enum Compression {
+pub enum Level {
     /// Fastest quality of compression, usually produces bigger size.
     Fastest,
     /// Best quality of compression, usually produces the smallest size.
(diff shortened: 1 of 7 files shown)


Nemo157:
One option would be to just have a constructor `pub fn with_quality(reader: $inner, level: u32) -> Self` and leave designing the parameters/builder interface till later. That gets the most common parameter available and will be backwards compatible with introducing a `with_params` or configuration methods later.

daxpedda:
Like [this](https://github.com/Nemo157/async-compression/pull/73/files#diff-e1576943a5907b15388ab726037ebdb3R24-R40)?

What about all the others?:
-  `BzEncoder` doesn't have a quality level, just `Default`, `Fastest`, `Best`, we could just do a 0 to 2 here
-  all the others have similar options, but additionally a level, so we can just copy paste the API from `BrotliEncoder`, that would make the other options unavailable for now

Nemo157:
Hmm, actually, looking at them all together I'm tempted to have a lowest-common-denominator API based on `BzEncoder`'s, just have a

```rust
pub enum Compression {
    Fast,
    Best,
    Default,
}
```

at the root, and have all encoders take that in with `pub fn with_quality(reader: $inner, level: Compression) -> Self`. Then translate that into the encoder specific parameters internally. That likely covers 99%+ of usecases, if you're really heavily optimizing the specific encoding parameters you're unlikely to be using async-io along with it.

daxpedda:
Done.

src/lib.rs
```diff
@@ -131,28 +131,38 @@ pub mod futures;
...
+mod util;
 
-/// Types to configure [`brotli`](::brotli) based encoders.
-#[cfg(feature = "brotli")]
-#[cfg_attr(docsrs, doc(cfg(feature = "brotli")))]
-pub mod brotli {
-    pub use brotli::enc::backward_references::BrotliEncoderParams;
+/// Level of compression data should be compressed with.
+#[derive(Clone, Copy, Debug)]
+pub enum Compression {
```
daxpedda: I put it in the crate root, is that okay?

src/macros.rs
```diff
@@ -21,70 +21,130 @@ macro_rules! algos {
...
+                        crate::codec::BzEncoder::new(bzip2::Compression::Default, 0),
+                    ),
+                }
+            }
+        } {
+            pub fn with_quality(inner: $inner, level: crate::Compression) -> Self {
+                Self {
+                    inner: crate::$($mod::)+generic::Encoder::new(
+                        inner,
+                        crate::codec::BzEncoder::new(level.into(), 0),
```
daxpedda: I'm assuming we don't want to change the `work_factor` here, even for `Best` compression quality?

src/macros.rs
```diff
@@ -21,14 +21,24 @@ macro_rules! algos {
...
+                Self {
+                    inner: crate::$($mod::)+generic::Encoder::new(
+                        reader,
+                        crate::codec::BrotliEncoder::new(params),
+                    ),
+                }
             }
         } {
-            pub fn from_params(inner: $inner, params: brotli::enc::backward_references::BrotliEncoderParams) -> Self {
+            /// The `level` argument here is typically 0-11.
```
Nemo157: This comment is out of date, can probably just drop it.

src/macros.rs
```diff
@@ -39,52 +49,102 @@ macro_rules! algos {
...
                         inner,
-                        crate::codec::ZlibEncoder::new(level),
+                        crate::codec::ZlibEncoder::new(level.into()),
                     ),
                 }
             }
         });
 
         algos!(@algo zstd ["zstd"] ZstdDecoder ZstdEncoder<$inner> {
             /// The `level` argument here can range from 1-21. A level of `0` will use zstd's default, which is `3`.
```
Nemo157: This comment as well.

src/macros.rs
```diff
@@ -21,14 +21,24 @@ macro_rules! algos {
 
     ($($mod:ident)::+<$inner:ident>) => {
         algos!(@algo brotli ["brotli"] BrotliDecoder BrotliEncoder<$inner> {
-            /// The `level` argument here is typically 0-11.
-            pub fn new(reader: $inner, level: u32) -> Self {
-                let mut params = brotli::enc::backward_references::BrotliEncoderParams::default();
-                params.quality = level as _;
-                Self::from_params(reader, params)
+            pub fn new(reader: $inner) -> Self {
+                let params = brotli::enc::backward_references::BrotliEncoderParams::default();
```
Nemo157: You could replace the body of `new` with `Self::with_quality(reader, crate::Compression::Default)`, that could also let you move it to the above macro branch to avoid repeating it for each algorithm

```rust
            $encoder<$inner> {
                pub fn new(inner: $inner) -> Self {
                    Self::with_quality(inner, crate::Compression::Default)
                }
            } $({ $($constructor)* })*
```

src/lib.rs
Nemo157: Crate root is fine, either name seems fine to me, up to you which to use.

src/macros.rs
daxpedda: That's great, done!

src/lib.rs
```diff
@@ -131,28 +131,38 @@ pub mod futures;
...
+    /// Default quality of compression defined by the selected compression algorithm.
+    Default,
 }
 
-/// Types to configure [`bzip2`](::bzip2) based encoders.
 #[cfg(feature = "bzip2")]
-#[cfg_attr(docsrs, doc(cfg(feature = "bzip2")))]
-pub mod bzip2 {
-    pub use bzip2::Compression;
+impl From<Compression> for bzip2::Compression {
```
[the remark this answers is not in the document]
daxpedda: Done, I also did that for all the other types, even if just converting to `i32`, just to have everything at one place.

src/lib.rs
daxpedda: I named it `Level`? I hope that's okay.

Nemo157:
Awesome, thanks a lot :)

bors r+

PR #76, opened 2020-02-11
Add opt-in support for decoding multi-member compressed streams

### LINKED ISSUE #75: Handle concatenated `.gz` files
I wanted to process a huge `.gz` file without downloading it which actually was concatenation of multiple `.gz` files. This is rather common practice to be able to utilzie multiple cores while compressing big amount of data.

`GzipDecoder` failed to handle it properly - it stopped after the first "chunk" and didn't even output an error.

`gzip` utility decompresses the file without any problems. I guess, so should do `GzipDecoder`.

Nemo157:
fixes #75 
closes #74

### DIFF
--- a/src/codec/brotli/decoder.rs
+++ b/src/codec/brotli/decoder.rs
@@ -56,6 +56,15 @@ impl BrotliDecoder {
 }
 
 impl Decode for BrotliDecoder {
+    fn reinit(&mut self) -> Result<()> {
+        self.state = BrotliState::new(
+            StandardAlloc::default(),
+            StandardAlloc::default(),
+            StandardAlloc::default(),
+        );
+        Ok(())
+    }
+
     fn decode(
         &mut self,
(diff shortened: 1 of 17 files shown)


Nemo157:
@bekh6ex what do you think of this as a solution for #75, it should handle concatenated GZip streams with an opt-in by the user.

alecmocatta:
Thanks for adding this @Nemo157! I have the same use case as @jpiepkow in the issue thread (processing the common crawl dataset) and I can confirm this PR solves it – see the example [here](https://github.com/alecmocatta/reqwest_resume/#Example).

I also agree that making this behaviour configurable rather than hardcoded as proposed in the issue thread is strongly desirable.

Is there a reason for supporting bufread but not stream?

Nemo157:
@alecmocatta thanks for trying it out :smile:

IIRC there's no reason it supports only bufread, I assume stream support will be easy to add.

I'll finish off the rebase to support the new formats and add stream support soon.

PR #81, opened 2020-03-08
add support for lzma

PaulAvery:
Hi, I attempted to add lzma support to your library as I needed it for my own usecase.

I hope I caught all the areas (tests, docs, etc) in which I had to add support for a new encoder/decoder, if not please let me know.

### DIFF
--- a/src/codec/lzma/encoder.rs
+++ b/src/codec/lzma/encoder.rs
@@ -1,0 +1,96 @@
+use crate::{codec::Encode, util::PartialBuffer};
+
+use std::fmt::{Debug, Formatter, Result as FmtResult};
+use std::io::Result;
+use xz2::stream::{Action, Check, Status, Stream};
+
+pub struct LzmaEncoder {
+    stream: Stream,
+}
+
+impl Debug for LzmaEncoder {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        write!(f, "LzmaEncoder")
+    }
+}
+
+impl LzmaEncoder {
+    pub fn new(level: u32) -> Self {
+        Self {
+            stream: Stream::new_easy_encoder(level, Check::Crc64).unwrap(),
+        }
+    }
+}
+
+impl Encode for LzmaEncoder {
+    fn encode(
+        &mut self,
+        input: &mut PartialBuffer<impl AsRef<[u8]>>,
+        output: &mut PartialBuffer<impl AsRef<[u8]> + AsMut<[u8]>>,
+    ) -> Result<()> {
+        let previous_in = self.stream.total_in() as usize;
+        let previous_out = self.stream.total_out() as usize;
+
+        let status = self
+            .stream
+            .process(input.unwritten(), output.unwritten_mut(), Action::Run)?;
+
+        input.advance(self.stream.total_in() as usize - previous_in);
+        output.advance(self.stream.total_out() as usize - previous_out);
+
+        match status {
+            Status::Ok | Status::StreamEnd => Ok(()),
+            Status::GetCheck => panic!("Unexpected lzma integrity check"),
+            Status::MemNeeded => Err(std::io::Error::new(
+                std::io::ErrorKind::Other,
+                "out of memory",
+            )),
+        }
+    }
+
+    fn flush(
+        &mut self,
+        output: &mut PartialBuffer<impl AsRef<[u8]> + AsMut<[u8]>>,
+    ) -> Result<bool> {
+        let previous_out = self.stream.total_out() as usize;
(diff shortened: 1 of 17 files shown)


Nemo157:
Awesome, looks good to me at first glance, I think you got everything that's needed. I'll have time tomorrow to go over it properly.

src/codec/lzma/encoder.rs
```diff
@@ -0,0 +1,96 @@
...
+impl Debug for LzmaEncoder {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        write!(f, "LzmaEncoder")
+    }
+}
+
+impl LzmaEncoder {
+    pub fn new(level: u32) -> Self {
+        Self {
+            stream: Stream::new_easy_encoder(level, Check::Crc64).unwrap(),
```
Nemo157: This and the decoder appear to be using [the XZ file format](https://tukaani.org/xz/xz-file-format.txt) rather than raw LZMA. Should they be called `XzDecoder`/`XzEncoder` to indicate that, or did you mean to actually have raw LZMA here? (Potentially it would make sense to support all of LZMA, LZMA2 and XZ formats as separate encoder/decoder pairs).

PaulAvery: You are right, they should probably be called `XzDecoder`/`XzEncoder`. I thought `lzma` and `xz` could be used interchangeably, but i was wrong. Then again I am not sure what an LZMA2 decoder/encoder would do, as the en-/decoding process seems to only make sense in the context of a specific file format? So we could do `LzmaEncoder` for `.lzma` files and `XzEncoder` for `.xz` files?
Or am I still misunderstanding how the LZMA/LZMA2 interact with the various file formats?

Nemo157: It would be possible to use LZMA2 as part of some other kind of data encoding, e.g. the `xz` format itself wraps an LZMA2 data stream. But by the looks of it `xz2` doesn't support encoding directly to LZMA2, `new_easy_encoder` gives you an `xz` encoder, `new_lzma_encoder` gives you an `lzma` encoder and `new_stream_encoder` gives you a customised `xz` encoder.

So for now adding separate `LzmaEncoder` and `XzEncoder` would be great (under separate `xz` and `lzma` features, probably sharing a base `Xz2Encoder`, similar to how `zlib` and `deflate` share most of their implementation).

PaulAvery: Ok, I started with what you suggested. Will have to add tests for the new LzmaEncoder/Decoder later today or tomorrorw.

src/codec/mod.rs
```diff
@@ -11,6 +11,12 @@ mod deflate;
 mod flate;
 #[cfg(feature = "gzip")]
 mod gzip;
+#[cfg(feature = "lzma")]
+mod lzma;
+#[cfg(feature = "xz")]
+mod xz;
+#[cfg(any(feature = "xz", feature = "lzma"))]
```
Nemo157: You can use `cfg(feature = "xz2")` here to simplify it, optional crates are also feature names.

tests/proptest.rs
```diff
@@ -76,4 +76,4 @@ macro_rules! tests {
     }
 }
 
-tests!(brotli, bzip2, deflate, gzip, zlib, zstd);
+tests!(brotli, bzip2, deflate, gzip, zlib, zstd, xz);
```
Nemo157: Should have `lzma` added as well.

src/codec/xz2/encoder.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    fn flush(
+        &mut self,
+        output: &mut PartialBuffer<impl AsRef<[u8]> + AsMut<[u8]>>,
+    ) -> Result<bool> {
+        let previous_out = self.stream.total_out() as usize;
+
+        let status = self
+            .stream
+            .process(&[], output.unwritten_mut(), Action::SyncFlush)?;
```
Nemo157: I think `SyncFlush` is unsupported for the `lzma` encoder (which is causing the test failures), should probably have `LzmaEncoder::flush` be a noop.

Nemo157: (based on the [`Stream::new_lzma_encoder`](https://docs.rs/xz2/0.1.6/xz2/stream/struct.Stream.html#method.new_lzma_encoder) docs).

src/codec/mod.rs
PaulAvery: Huh, I did not know that. Thanks :)

Nemo157:
Looks good, thanks a lot!

bors r+

PR #86, opened 2020-04-20
RFC: Add Level::Precise to allow precise quality level specification.

lseelenbinder:
First thanks for making this! We're very happily migrating some sync code to use this within a hyper server.

Since we're using this crate in a high-performance speed and size sensitive web service, we need to tune our underlying algorithm qualities beyond just `Fastest`, `Best` and `Default`. For example, Brotli's default value is the same as best, which is much too slow for us, but `Fastest` is basically a no-op when measuring compression ratios.

This PR exposes a `Level::Precise(u32)` option to allow specific tuning of the underlying encoder's quality; this works for all algorithms except bzip2, where it just uses `Default`. In basic testing it works like a charm.

I was unable to build tests on latest `master`, so I'm not sure how to begin testing.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -162,6 +162,10 @@ pub enum Level {
     Best,
     /// Default quality of compression defined by the selected compression algorithm.
     Default,
+    /// Precise quality based on the underlying compression algorithms'
+    /// qualities. The interpretation of this depends on the algorithm choosen.
+    /// For bzip2, this is treated as default.
+    Precise(u32),
 }
 
 impl Level {
@@ -170,6 +174,7 @@ impl Level {
         match self {
             Self::Fastest => params.quality = 0,
             Self::Best => params.quality = 11,
+            Self::Precise(quality) => params.quality = quality as i32,
             Self::Default => (),
         }
 
@@ -181,6 +186,7 @@ impl Level {
         match self {
             Self::Fastest => bzip2::Compression::Fastest,
             Self::Best => bzip2::Compression::Best,
+            Self::Precise(_) => bzip2::Compression::Default,
             Self::Default => bzip2::Compression::Default,
         }
     }
@@ -190,6 +196,7 @@ impl Level {
         match self {
             Self::Fastest => flate2::Compression::fast(),
             Self::Best => flate2::Compression::best(),
+            Self::Precise(quality) => flate2::Compression::new(quality as u32),
             Self::Default => flate2::Compression::default(),
         }
     }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -162,6 +162,10 @@ pub enum Level {
     Best,
     /// Default quality of compression defined by the selected compression algorithm.
     Default,
+    /// Precise quality based on the underlying compression algorithms'
+    /// qualities. The interpretation of this depends on the algorithm choosen.
+    /// For bzip2, this is treated as default.
```
Nemo157, line 167: I think the special case for bzip2 doesn't really need to be mentioned, it's implicit in this value being implementation specific anyway.

I also think that the fact that this is _implementation_ specific, not just _algorithm_ specific should be highlighted. The main reason I want this local enum instead of just re-exporting the underlying configuration from the different implementations is to allow switching implementations of algorithms in the future (e.g. I would really love to drop all the C dependencies at some point when viable Rust replacements are available).

```suggestion
    /// qualities. The interpretation of this depends on the algorithm chosen
    /// and the specific implementation backing it.
```

Nemo157:
Overall I do think this is a good first step towards supporting more configuration, for most algorithms there are pretty well known "levels" that are supported across different implementations.

One thing I do wonder is what the libraries do if the level passed is outside their range, should the conversion implicitly clamp them, or is this handled well internally?

lseelenbinder:
> Overall I do think this is a good first step towards supporting more configuration, for most algorithms there are pretty well known "levels" that are supported across different implementations.

👍 

> One thing I do wonder is what the libraries do if the level passed is outside their range, should the conversion implicitly clamp them, or is this handled well internally?

I meant to highlight this in my initial comment. I think this should be addressed before this change is merged. Either by documenting choosing a value outside the range may result in a panic or undefined behavior or by enforcing the limits ourselves.

Nemo157:
I think my preference would be to implicitly clamp the values. Fundamentally that's what the `bzip2` implementation is doing by mapping it to the default.

lseelenbinder:
> I think my preference would be to implicitly clamp the values.

Sounds good.

I can think of a few ways to accomplish this; what's your preference?

1. `max([input], [known maximum for algorithm])`
2. Error on `Precise(quality) => quality > [known maximum]` (this would be similar to a `try_into` implementation).
3. Use a concrete `PreciseLevel` enum to restrict it to N levels.

I prefer 1.

Nemo157:
1 sounds good

lseelenbinder:
Implemented and updated documentation to reflect the fact we clamp the values to maximum of the underlying algorithm's maximum.

Nemo157:
Sorry about the delay, all LGTM, thanks!

bors r+

PR #91, opened 2020-05-15
Call inner writer's poll_flush instead of poll_close

gnosek:


gnosek:
Looks like a copy/paste error and with this patch my tests (several layers up) pass, but I'm curious to know if this is by design.

cc @Nemo157

Nemo157:
Definitely looks like a copy-paste error to me. Looks like I need to improve the writer used in the tests to actually close itself when `poll_close` is called so that it can detect this.

gnosek:
Thanks!

Nemo157:
I fixed the decoder as well in #92 and just published 0.3.4 with the fixes.

PR #95, opened 2020-06-27
Add support for tokio v0.2

### LINKED ISSUE #57: Add tokio-io trait support
Add `tokio::{bufread, write}` modules based on traits from `tokio-io 0.2`. Until that has a non-alpha release the modules should be extra gated behind some unstable flag.

Nemo157:
Terribad copy-paste job, but any cleanup to reduce duplication shouldn't affect public API so easy path is to get this in first.

fixes #57


Nemo157:
bors retry

PR #102, opened 2020-08-23
Add Examples

CfirTsabari:
Example  of zlip tokio write ecnoder and decoder.
Example reading zstd from stdin and write gzip to stdout, using tokio

Related to Nemo157/async-compression#101

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,7 @@ futures-test = "0.3.5"
 ntest = "0.3.3"
 timebomb = "0.1.2"
 bytes = "0.5.0"
-tokio-02 = { package = "tokio", version = "0.2.21", default-features = false, features = ["io-util", "stream"] }
+tokio-02 = { package = "tokio", version = "0.2.21", default-features = false, features = ["io-util", "stream", "macros", "io-std"] }
 tokio-03 = { package = "tokio", version = "0.3.0", default-features = false, features = ["io-util", "stream"] }
 tokio-util = { version = "0.4.0", default-features = false, features = ["io"] }
 
@@ -96,3 +96,11 @@ required-features = ["zlib"]
 [[test]]
 name = "zstd"
 required-features = ["zstd"]
+
+[[example]]
+name = "zlib_tokio_02_write"
+required-features = ["zlib", "tokio-02"]
+
+[[example]]
+name = "zstd_gzip"
+required-features = ["zstd", "gzip", "tokio-02"]
--- a/examples/zlib_tokio_02_write.rs
+++ b/examples/zlib_tokio_02_write.rs
@@ -1,0 +1,30 @@
+use async_compression::tokio_02::write::ZlibDecoder;
+use async_compression::tokio_02::write::ZlibEncoder;
+
+use std::io::Result;
+use tokio_02::io::AsyncWriteExt as _; // for `write_all` and `shutdown`
+
+use tokio_02 as tokio; // this enable the tokio main macro
+#[tokio_02::main]
+async fn main() -> Result<()> {
+    let data = b"example";
+    let compressed_data = compress(data).await?;
+    let de_compressed_data = decompress(&compressed_data).await?;
+    assert_eq!(de_compressed_data, data);
+    println!("{:?}", String::from_utf8(de_compressed_data).unwrap());
+    Ok(())
+}
+
+async fn compress(in_data: &[u8]) -> Result<Vec<u8>> {
+    let mut encoder = ZlibEncoder::new(Vec::new());
+    encoder.write_all(in_data).await?;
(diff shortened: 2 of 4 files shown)


examples/zlib_tokio_02_write.rs
```diff
@@ -0,0 +1,31 @@
+use async_compression::tokio_02::write::ZlibDecoder;
+use async_compression::tokio_02::write::ZlibEncoder;
+
+use std::io::Result;
+use tokio_02::io::AsyncWriteExt; // for `write_all` and `shutdown`
```
Nemo157, line 5:
```suggestion
use tokio_02::io::AsyncWriteExt as _; // for `write_all` and `shutdown`
```

examples/zlib_tokio_02_write.rs
```diff
@@ -0,0 +1,31 @@
...
+use tokio_02::runtime::Runtime;
+
+async fn async_main() {
```
Nemo157, line 8: Can you add the `macros` feature to the `tokio` dependency, and use `?` instead of `unwrap` everywhere

```suggestion
#[tokio::main]
async fn main() -> std::io::Result<()> {
```
async fn async_main() {
```

File: Cargo.toml [hunk 3]
```diff
@@ -93,3 +93,7 @@ required-features = ["zlib"]
 [[test]]
 name = "zstd"
 required-features = ["zstd"]
+
+[[example]]
+name = "zlib_tokio_02_write"
+required-features = ["zlib", "tokio-02"]
```
Nemo157, line 99: Should be a newline at the end of file.

Nemo157 (changes requested):
Hmm, my biggest issue with this is that it's showing something you really _should not_ do (memory -> memory (de)compression using `async`).

What do you think of instead having something that reads from `stdin` to memory (using `tokio_02::bufread::ZstdDecoder` wrapped around `tokio::io::Stdin`), logs the total length to `stderr`, then writes it back to `stdout` (using `tokio_02::write::GzipEncoder` wrapped around `tokio::io::Stdout`). That would show usage of both the `bufread` and `write` adaptors, and would be a much more realistic example.

Having a short comment showing how to run it would also be useful then (I chose zstd + gzip above to make it easy to test with cli tools):

```console
echo 'hello' | zstd | cargo run --example ... | gunzip -c
```

I've also left some inline comments on things I'd change to make this follow "best practices".

examples/zlib_tokio_02_write.rs
CfirTsabari: this will not work if I have 2 items like that

examples/zlib_tokio_02_write.rs
CfirTsabari: will fix

Cargo.toml
CfirTsabari: will fix

CfirTsabari:
> Hmm, my biggest issue with this is that it's showing something you really _should not_ do (memory -> memory (de)compression using `async`).
> 
> What do you think of instead having something that reads from `stdin` to memory (using `tokio_02::bufread::ZstdDecoder` wrapped around `tokio::io::Stdin`), logs the total length to `stderr`, then writes it back to `stdout` (using `tokio_02::write::GzipEncoder` wrapped around `tokio::io::Stdout`). That would show usage of both the `bufread` and `write` adaptors, and would be a much more realistic example.
> 
> Having a short comment showing how to run it would also be useful then (I chose zstd + gzip above to make it easy to test with cli tools):
> 
> ```
> echo 'hello' | zstd | cargo run --example ... | gunzip -c
> ```
> 
> I've also left some inline comments on things I'd change to make this follow "best practices".

I have added an example as you suggested.
But I also think that when it comes to examples sometime simplicity should overcome it's "not good practice".
And it's very common to see async examples that use memory vector.
That's why I think you should consider keeping both examples, (with a comment that warn about it).
As a newcomer, the first example helps me a lot more to understand how to work with `aync-compression`.

Of course, if you want me to I will remove it.

examples/zlib_tokio_02_write.rs
Nemo157: You can have multiple things renamed to `_`.

CfirTsabari: oh my bad it was an editor false negative that made me think I can't.
fixed it

examples/zlib_tokio_02_write.rs
CfirTsabari: fixed it

Cargo.toml
CfirTsabari: fixed it

Nemo157:
Sorry about the delay, rebased to fix a conflict with the `Cargo.toml`

bors r+

PR #104, opened 2020-10-18
Tokio 0.3 support

Nemo157:


Nemo157:
Docs appear to be broken because of <https://github.com/rust-lang/rust/issues/76296>.

PR #132, opened 2021-11-12
Disable default-features on flate2 to allow use of alternate backends

joshtriplett:


Nemo157:
You will have to add `flate2` into the dev-dependencies with default features active so the tests can use the default backend.

joshtriplett:
Ah, sorry. flate2 with default-features doesn't actually prevent the use of zlib-ng, it just *also* builds in the default miniz_oxide backend, which takes up space but doesn't otherwise do any harm. And while I could make flate2 avoid building any backend by default, that would mean the user of async-compression would have to pick a backend and would get similar errors if they did not. So, I'll just close this; it's not worth the usability hit just to avoid building (a potentially unused copy of) the default miniz_oxide backend.

PR #139, opened 2022-01-26
Use only `std` feature for `brotli`

mati865:
Closes https://github.com/Nemo157/async-compression/issues/83 by disabling `ffi-api` feature thanks to https://github.com/dropbox/rust-brotli/pull/48


Nemo157:
Nice, I hadn't realised that was actually released.

bors r+

PR #143, opened 2022-04-12
Update Cargo.toml

hariria:
Upgrading zstd-safe to 5.0.1+zstd.1.5.2

Passes all tests in 
```
cargo test --all-features
```


hariria:
Can we re-run on the newer changes?

hariria:
Can we re-run on the newer changes?

PR #148, opened 2022-05-12
bufread::generic::Decoder: don't reinitialize on reader EOF

eeeeeta:
If `multiple_members` is enabled, the `bufread::generic::Decoder` will
attempt to reinitialise the decoder inside `State::Flushing`, even if
the reason it entered that state was due to the reader returning an EOF.
This will result in an attempt to read past EOF, which is highly
undesirable to say the least.

To fix this, force `multiple_members` to `false` when we get an EOF
condition from the reader.

### DIFF
--- a/src/futures/bufread/generic/decoder.rs
+++ b/src/futures/bufread/generic/decoder.rs
@@ -70,6 +70,9 @@ impl<R: AsyncBufRead, D: Decode> Decoder<R, D> {
                 State::Decoding => {
                     let input = ready!(this.reader.as_mut().poll_fill_buf(cx))?;
                     if input.is_empty() {
+                        // Avoid attempting to reinitialise the decoder if the reader
+                        // has returned EOF.
+                        *this.multiple_members = false;
                         State::Flushing
                     } else {
                         let mut input = PartialBuffer::new(input);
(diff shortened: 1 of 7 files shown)


Nemo157:
To understand the issue I added an additional test assertion that I believe covers what this is meant to fix: https://github.com/Nemo157/async-compression/commit/bd6f0aea785852317b51b56ab60b6d3a9e392c01. If that looks like what you saw you can cherry-pick it to your branch.

eeeeeta:
@Nemo157 Yeah, that test looks good! Cherry-picked (and fixed the stupid compilation error :p)

src/futures/bufread/generic/decoder.rs
```diff
@@ -70,6 +70,9 @@ impl<R: AsyncBufRead, D: Decode> Decoder<R, D> {
                 State::Decoding => {
                     let input = ready!(this.reader.as_mut().poll_fill_buf(cx))?;
                     if input.is_empty() {
+                        // Avoid attempting to reinitialise the decoder if the reader
+                        // has returned EOF.
+                        *this.multiple_members = false;
```
Nemo157, line 75: This same change will also need to be applied to `src/{tokio,tokio_02,tokio_03}/bufread/generic/decoder.rs`.

eeeeeta:
Thanks, @Nemo157! Would you mind cutting a new release and pushing it to crates.io? (we're using your crate in [arti](https://gitlab.torproject.org/tpo/core/arti), and not having a git dependency would be nice!)

Nemo157:
Published 0.3.14

Issue #150, opened 2022-05-20
proposal: Export codec and `Encode`/`Decode` trait


Xuanwo:
We are using our own runtime in [databend](https://github.com/datafuselabs/databend/) which will:

- Running async task on async runtime (tokio here), for example, IO task.
- Running sync task on sync runtime, for example, CPU bound task like decompress.

So we want to control the underlying behavior of `async-compression`:

- Make IO happen on async runtime: poll the futures
- Make `decode` happen on sync runtime

Thus, we have to access the `Encode`/`Decode` trait in `codec` directly and build our own `XzDecoder` and so on.

Is this idea make sense to you? I'm willing to start a PR for it.

Xuanwo:
Maybe expose `decode` in `XzDecoder`?

```rust
impl<R: AsyncBufRead, D: Decode> Decoder<R, D> {
    pub fn decode(
        self: Pin<&mut Self>,
        input: &mut PartialBuffer<impl AsRef<[u8]>>,
        output: &mut PartialBuffer<impl AsRef<[u8]> + AsMut<[u8]>>
    ) -> Result<bool> {
        self.decoder.decode(&mut input, output)
    }
}
```

Nemo157:
I am (slowly) working on a big refactor that will expose a lot of the internal building blocks, including the `codec` stuff. I'm hoping to have enough time to actually get something public in the next few weeks.

Xuanwo:
That's cool! If there is anything I can help with, just ping me back.

Nemo157:
@Xuanwo I was thinking about your usecase a bit more, and it seems like you would need the codec traits to be async; so you can spawn the synchronous work into your sync-runtime and return a handle to get the result?

Xuanwo:
> so you can spawn the synchronous work into your sync-runtime and return a handle to get the result?

Yep, similar. In databend, we have our own processors like the following: 

https://github.com/datafuselabs/databend/blob/2f9e425b55e2e28761c5a9b1e3fb11131fd7fa8e/query/src/pipelines/new/processors/processor.rs#L25-L49

```rust
pub enum Event {
    NeedData,
    NeedConsume,
    Sync,
    Async,
    Finished,
}


// The design is inspired by ClickHouse processors
#[async_trait::async_trait]
pub trait Processor: Send {
    fn name(&self) -> &'static str;


    fn event(&mut self) -> Result<Event>;


    // Synchronous work.
    fn process(&mut self) -> Result<()> {
[...]

Issue #133, opened 2021-11-13
Please support enabling zstd long-distance matching mode


joshtriplett:
async-compression only allows passing a numeric level. I'd love to enable the zstd "long-distance matching mode", which substantially improves compression on large files that may have similarity at long distances.

I'd be happy to supply a PR implementing this. Would you prefer to extend `Level` to contain internal flags for things like this, or would you prefer to add a method (or additional constructor) directly to the `ZstdEncoder` to set this parameter?

joshtriplett:
I submitted https://github.com/Nemo157/async-compression/pull/134 implementing this. (Happy to take a different approach if you would prefer.)

Nemo157:
Unfortunately I want to keep the algorithmic dependencies of `async-compression` private to allow non-breaking updates.

I have long been meaning to split out the underlying generic implementation that exposes these and would allow full configurability for different libraries, with the downside of needing to either support multiple implementations/versions of the algorithms or doing breaking releases.

I think I had started an attempt at that a while ago, but found it less trivial than I was hoping. This is some motivation to look at it again, hopefully I can at least see how difficult it appears this weekend.

joshtriplett:
Would you consider a PR adding an abstraction over a handful of the options, to avoid exposing the underlying library? I'd be happy to prepare such a patch.

joshtriplett:
This is now fixed, with #134.

PR #152, opened 2022-06-26
Remove support for tokio 0.2 and 0.3

Nemo157:
I have checked the index using `rg -IN 'async-compression' | jq -r '"\(.name) \(.vers): \(.deps[] | select(.name == "async-compression") | .features | join(", "))"'` and there are only 2 crates that have their latest version still using `tokio-02` and no crates using `tokio-03`.

I wanted to wait for `std` to introduce some async IO traits and go directly to 1.0 when removing this support, but given how long it has been and how far away that still appears to be, I think it's time to do this now. This will simplify exposing more of the internals in order to allow some of the usecases people want (#133, #150).


Xuanwo:
Nice change!

Xuanwo:
@Nemo157 Is there other concerns that block this PR get merged?

robjtede (approved):

Nemo157:
I think it would be worth merging a zstd update and pushing that before going to 0.4, so that there's less pressure on downstream updating.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "async-compression"
-version = "0.3.15"
+version = "0.4.0"
```
robjtede, line 3: The script I mentioned in #200 can now keep this in sync with the changelog.

Nemo157: Ok, I'll drop that commit and rebase over #200 then.

Issue #154, opened 2022-08-24
Encoder does not support intermediate flushes before the reader has ended


Geal:
*disclaimer: this is about fixing https://github.com/apollographql/router/issues/1572, for which I'm under tight timing constraints, so right now this issue's goal (and the related PR) is discussing and finding a quick fix that we can use directly in the router, and then I'll help find a proper solution.*

I'm encountering an issue with the following setup:
- I use axum with tower-http's `CompressionLayer`, which uses async-compression
- the router sends a HTTP response with multipart
- one multipart element is sent right away, the next one is sent after 5s
- if the response is compressed, the router will wait for 5s to send both parts at the same time, while without compression, the first part comes immediately, and the second one after 5s

I tracked that down to async-compression, where in the tokio based `Encoder`, due to the use of `ready!()`, whenever the underlying reader returns `Poll::Pending`, it is transmitted directly to the caller, so there is no flush until the entire data has been read:

https://github.com/Nemo157/async-compression/blob/ada65c660bcea83dc6a0c3d6149e5fbcd039f739/src/tokio/bufread/generic/encoder.rs#L63-L74

I would like the encoder to send the data it already compressed when the reader returned `Poll::Pending`, and let the upper layer decide on buffering or sending the HTTP chunk directly.

Nemo157:
It sounds like there's 2 parts to this:

1. similar to #123 if we have data available from the encoder, but the input isn't ready, we won't send that data out
2. you also need to be able to force the encoder to flush any stored data at the end of your first multipart element

For the second part, there's no real way to indicate that from the `BufRead`. Flushing any time a `Pending` is received will reduce the encoding efficiency since the encoders generally build up state across multiple input chunks to reduce overhead, a flush forces it to output everything it has internally buffered so far.

This would be easy to do in an `impl Write` based API

```rust
response.write_all(first_element).await?;
response.flush().await?;
response.write_all(second_element).await?;
```

I'm not sure how to support it for `BufRead`.

Geal:
right, `BufRead` does not really support such hints. I'm in a special case here, where I know exactly when I have enough data to send, and must send it as soon as possible, so I don't care much about compression efficiency, but that's not the general case. Although there should be a way to flush from time to time, otherwise we'll still have the entire compresed data stored in a buffer before sending, which defeats the purpose of a streaming API

Nemo157:
Maybe an option to enable this behavior would be the best approach `fn flush_on_pending(&mut self, bool)`

[...]

PR #155, opened 2022-08-24
support intermediate flushes when encoding

Geal:
*disclaimer: this is an attempt at fixing https://github.com/apollographql/router/issues/1572, for which I'm under tight timing constraints, so right now this PR's goal is discussing and finding a quick fix that we can use directly in the router, and then I'll help find a proper solution.*

tentative fix for #154

Here I make sure to call the `flush` method on pending, but what happens here is that one 64 bytes chunk of compressed data is returned immediately (not enough for the first part of the encoded data), and then it waits for 5s for the rest. Any advice on making it work? Does it need to flush multiple times?

### DIFF
--- a/src/codec/flate/encoder.rs
+++ b/src/codec/flate/encoder.rs
@@ -67,18 +67,6 @@ impl Encode for FlateEncoder {
             FlushCompress::Sync,
         )?;
 
-        loop {
-            let old_len = output.written().len();
-            self.encode(
-                &mut PartialBuffer::new(&[][..]),
-                output,
-                FlushCompress::None,
-            )?;
-            if output.written().len() == old_len {
-                break;
-            }
-        }
-
         self.flushed = true;
         Ok(!output.unwritten().is_empty())
     }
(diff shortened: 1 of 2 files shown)


jeromegn:
I think this effort has been abandoned (because the apollo graphql router just removed compression for multipart responses), but we're facing a similar issue with "html streaming" where the content-type is text/html. It's a slow-drip stream too.

Reproduction:

```
curl -N https://frosty-frog-2648.fly.dev/ --compressed
```

Trying this branch (thankfully) broke our tests. It looks like the read loop is stuck like:

```
encoding -> flushing ->  encoding -> flushing -> ...
```

Since it has already read some bytes, `read != 0`, and so it will never ever return `Poll::Pending` after that. I have a branch where I'm just resetting `read = 0` when I've flushed successfully. I don't know if that renders this fix useless or if I've broken something else.

Geal:
It's not abandoned, I'm just under tight constraints, due to moving out, and the router remease coming up soon 😅

I'd like to look a bit at using a writer oriented approach from inside tower-http, maybe that will be a cleaner way

PR #157, opened 2022-08-28
Add dependabot

NobodyXu:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+# Dependabot dependency version checks / updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    # Workflow files stored in the
+    # default location of `.github/workflows`
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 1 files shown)


Nemo157:
In my experience dependabot is not useful for libraries, and results in a lot of overhead as it creates PR after PR (I see far too many repos on github with dozens of open dependabot PRs).

There is one thing I think could be useful in this space, being notified on new _major_ versions of libraries; but that doesn't appear to be something dependabot supports.

NobodyXu:
I've used dependabot on other library crates such as [openssh](https://github.com/openssh-rust/openssh/pulls?q=is%3Apr+author%3Aapp%2Fdependabot) and it only generates a PR on new major ver of dependencies.

Nemo157:
Interesting, maybe they changed that behavior since merging into github.

Could you add [`rebase-strategy: "disabled"`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#rebase-strategy) as the unnecessary rebasing it does doesn't work well with bors; then I'll give it a test.

NobodyXu:
> Interesting, maybe they changed that behavior since merging into github.
> 
> Could you add [`rebase-strategy: "disabled"`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#rebase-strategy) as the unnecessary rebasing it does doesn't work well with bors; then I'll give it a test.

@Nemo157 Done!

PR #166, opened 2022-08-31
Delete Cargo.lock

NobodyXu:
`Cargo.lock` seems to cause dependabot to open new PR for every minor release.
Removing `Cargo.lock` should fix it, since a library crate won't need it anyway.


Nemo157:
I prefer to keep the `Cargo.lock` to support usecases like bisection that need a known good set of dependencies. I think to satisfy dependabot I'm gonna just keep telling it to ignore versions until it stops opening PRs... (afaict from the docs there is no way to configure it to only PR on major updates, that's only done via auto-detection).

NobodyXu:
> I think to satisfy dependabot I'm gonna just keep telling it to ignore versions until it stops opening PRs... (afaict from the docs there is no way to configure it to only PR on major updates, that's only done via auto-detection).

That's unfortunate.
If I find solution to this problem, I will open a PR here.

PR #178, opened 2022-10-17
Streaming encoder

### LINKED ISSUE #154: Encoder does not support intermediate flushes before the reader has ended

Geal:
Fix #154 

:warning: not ready yet, only the bare minimum was done to validate it in tokio and gzip, the rest wil come later

this is a follow up on the proposal for a `poll_flush` implementation. It is implemented on `tokio::bufread::Encoder` and in the macros. We then provide a `StreamingEncoder` that wraps a `AsyncRead` and `AsyncFlush` implementation to flush it if it produced nothing after a configurable time.

The test shows it appears  to work, but I'd like to test it more using the proptest integration.
On the API side, I think it will look better as a `.flush_timeout()` method?

I looked into flushing once a configurable amount of data has been passed, but that requires hooking the data producer to count the bytes passing through, pass it to the `Encoder`, then have the `StreamingEncoder` monitor that count. That's a bit involved, maybe that could be a separate implementation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -46,6 +46,7 @@ dependencies = [
  "tokio 0.2.25",
  "tokio 0.3.7",
  "tokio 1.15.0",
+ "tokio-stream",
  "tokio-util 0.3.1",
  "tokio-util 0.4.0",
  "tokio-util 0.5.1",
@@ -725,7 +726,7 @@ dependencies = [
  "memchr",
  "pin-project-lite 0.1.12",
  "slab",
- "tokio-macros",
+ "tokio-macros 0.2.6",
 ]
 
 [[package]]
@@ -750,6 +751,7 @@ dependencies = [
  "bytes 1.1.0",
  "memchr",
  "pin-project-lite 0.2.8",
+ "tokio-macros 1.8.0",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


Geal:
@Nemo157 I cleaned it up a bit, to make it testable with other algorithms than just gzip. Should I really add implementations for other targets than tokio? Considering that support for tokio 0.2 and 0.3 is removed in https://github.com/Nemo157/async-compression/pull/152, the `stream` feature is deprecated, and the `futures-io` target makes no assumption on the runtime, I only see tokio as a possible target here?
(this PR needs adherence to the runtime because it uses a timeout)

Another approach I'd like investigate is having the streaming encoder expose a handler of some kind (channel, whatever) that could be used to trigger a flush externally. The user would then decide when to flush, possibly based on timeouts, or other conditions (like the deferred responses in the graphql router). That would remove the link to the runtime, and let higher level libraries decide on the flushing strategy

Geal:
I removed the timeout based solutin, this should be enough to review now @Nemo157

Geal:
@Nemo157 could you activate the CI on  this PR?

Nemo157:
Hmm, reminds me I need to find out _why_ github has changed the default to not running CI, afaik it's supposed to be secure against malicious PRs :thinking: 

I'll hopefully have some time to review this week.

Geal:
@Nemo157 friendly ping? Is there any way I can make this easier to review for you?

Geal:
@Nemo157 looks like the CI is passing. Do you think anything is missing from this PR?

Geal:
@Nemo157 FYI I'll vendor part of the compression code in the router, so I can fix the bug on my end, so no pressure for you on this PR, review it when it's convenient :)

robjtede:
@NobodyXu since this is additive, I propose we review this after the 0.4.0 release ?

NobodyXu:
Agree, I can't see how this could result in a breaking change.

NobodyXu:
Plus it's not ready yet, so IMO we should go ahead and release 0.4.0

PR #199, opened 2023-05-08
Migrate from bors to merge queues

Nemo157:
Also needs reconfiguring in the settings, I'll do that if you want to use merge queues, the main upside is that it skips the slow exhaustive job before showing PR status, and that job is unlikely to be broken with most changes. Otherwise I could just update that job to run on pull-request too and have a normal one-click-merge.


robjtede (approved):

robjtede:
Nice, definitely prefer native queue over bors. If you already understand the settings page for it then please set it up so I can see how it's done.

robjtede:
In Actix Web we require up-to-date branches to merged tests pass but that's a pain for contributors when their PRs get out of date, so this method is great news to find out about.

Nemo157:
Ok, I hadn't realised that the required statuses must pass both _before_ and during the merge queue, and you can only configure a single set of required status checks. I guess I'll set the exhaustive tests to run both times.

PR #202, opened 2023-05-08
ci: cache dependencies in all exhaustive jobs

robjtede:


NobodyXu:
P.S. Maybe we can use [sccache-actuon](https://github.com/marketplace/actions/sccache-action) in other jobs as well to cache dependencies?

robjtede:
I don't think there's any need for it, in my experience this does a really good job.

NobodyXu (approved):

NobodyXu:
Looks good to me

PR #204, opened 2023-05-09
dependabot: only update deps on breaking changes

NobodyXu:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,7 @@ updates:
     rebase-strategy: "disabled"
   - package-ecosystem: "cargo"
     directory: "/"
+    versioning-strategy: "widen"
     schedule:
       interval: "daily"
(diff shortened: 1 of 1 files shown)


robjtede:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy

docs suggest that this isn't supported by the `cargo` updater

NobodyXu:
> https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy
> 
> docs suggest that this isn't supported by the `cargo` updater

It should be because this is the default strategy for crates without `Cargo.lock`.

I have other lib crates (without `Cargo.lock`) enabled dependabot and dependabot would only post a PR if one of the dep release a breaking change.

PR #217, opened 2023-05-10
Accept negative `Level::Precise`

### LINKED ISSUE #179: not all ZStd compression levels are usable
the quality level of zstd is mapped like this `quality.min(21) as i32`. ZStd goes up to level 22, which is unreachable.
Another issue that is probably a bit harder to fix is the fact that the compression levels of zstd start at -7, while `Level::Precise` uses an `u32` for representing levels.
maybe it would be appropriate to bump the major version and swap the `u32` with an `i32` (or maybe even an `i16`, i doubt level 10000 would be surpassed by any algorithm)

NobodyXu:
Fixed #179

Also, for bzip2, flate2 and zstd, use min/max level provided by the crate instead of hard-code these values into async-compression.


src/lib.rs
```diff
@@ -177,30 +180,49 @@ impl Level {
...
     #[cfg(feature = "zstd")]
     fn into_zstd(self) -> i32 {
+        let (fastest, best) = libzstd::compression_level_range().into_inner();
         match self {
-            Self::Fastest => 1,
-            Self::Best => 21,
-            Self::Precise(quality) => quality.min(21) as i32,
+            Self::Fastest => fastest,
+            Self::Best => best,
+            Self::Precise(quality) => quality.max(fastest).min(best),
```
robjtede, line 225:
```suggestion
            Self::Precise(quality) => quality.clamp(fastest, best),
```

src/lib.rs
```diff
@@ -177,30 +180,49 @@ impl Level {
...
-            Self::Best => flate2::Compression::best(),
-            Self::Precise(quality) => flate2::Compression::new(quality.min(10)),
+            Self::Fastest => fastest,
+            Self::Best => best,
+            Self::Precise(quality) => flate2::Compression::new(
+                quality
+                    .try_into()
+                    .unwrap_or(0)
+                    .max(fastest.level())
+                    .min(best.level()),
```
robjtede, line 213:
```suggestion
                    .clamp(fastest.level(), best.level())
```

src/lib.rs
```diff
@@ -177,30 +180,49 @@ impl Level {
...
-            Self::Best => bzip2::Compression::best(),
-            Self::Precise(quality) => bzip2::Compression::new(quality.max(1).min(9)),
+            Self::Fastest => fastest,
+            Self::Best => best,
+            Self::Precise(quality) => bzip2::Compression::new(
+                quality
+                    .try_into()
+                    .unwrap_or(0)
+                    .max(fastest.level())
+                    .min(best.level()),
```
robjtede, line 194: avoids the counter-intuitive max-min clamping

```suggestion
                    .clamp(fastest.level(), best.level())
```

src/lib.rs
robjtede, line 213:
```suggestion
                    .clamp(fastest.level(), best.level())
```

robjtede (changes requested):

robjtede (approved):
the public API change should have a changelog entry

NobodyXu:
> the public API change should have a changelog entry

done

robjtede (approved):

robjtede:
bypassing merge queue as GHA is having an outage

PR #225, opened 2023-05-14
Use `cargo-nextest` to speedup testing

robjtede:
Imo this is plenty fast enough and any further changes would only complicate our simple CI setup for no to negligible gain.

### DIFF
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -10,7 +10,10 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - uses: actions-rust-lang/setup-rust-toolchain@v1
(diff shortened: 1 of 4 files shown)


NobodyXu (changes requested):
Can we also disable incremental build and use sparse crates.io registries?

    CARGO_INCREMENTAL: 0
    CARGO_REGISTRIES_CRATES_IO_PROTOCOL: sparse

robjtede:
the toolchain setup action does those things

NobodyXu:
> the toolchain setup action does those things

oh you are right, this PR LGTM

NobodyXu (approved):

PR #229, opened 2023-06-21
Add support for zstd dictionaries

miam-miam100:
zstd dictionaries allow you to train on similar data to achieve better compression ratios on small files. This PR provides no way for people to train their own data, just use pre-trained dictionaries for decoding and encoding.


src/codec/zstd/decoder.rs
```diff
@@ -14,6 +14,13 @@ impl ZstdDecoder {
             decoder: Unshared::new(Decoder::new().unwrap()),
         }
     }
+
+    pub(crate) fn new_with_dict(dictionary: &[u8]) -> Self {
+        let mut decoder = Decoder::with_dictionary(dictionary).unwrap();
```
NobodyXu, line 19: loading dict could fail, so unwrapping here isn't a good idea.

src/macros.rs
```diff
@@ -98,8 +122,24 @@ macro_rules! algos {
             }
         });
 
-        algos!(@algo xz ["xz"] XzDecoder XzEncoder<$inner> {
-            pub fn with_quality(inner: $inner, level: crate::Level) -> Self {
+        algos!(@algo-dec zstd ["zstd"] ZstdDecoder<$inner_dec> {
```
NobodyXu, line 125: Why can't you write 

```suggestion
impl<R> ZstdDecoder<R> {
```

then you won't need to change the macros.

NobodyXu (changes requested):

src/codec/zstd/decoder.rs
miam-miam100: No problem been changed, is it guaranteed that `new` and `new_with_params` can't panic?

src/macros.rs
miam-miam100: Bcs then we would need to specify the trait R implements
```rs
impl<R: tokio::io::AsyncBufRead> ZstdDecoder<R> {
``` 
which is dependent on the async executor.

NobodyXu: Thanks, you are right in this

NobodyXu (approved):

NobodyXu:
@robjtede Can you review this please?
Since this involves changes to macro_rules, I'd like to review from another maintainer.

NobodyXu:
@robjtede Can you review this and get this into main before the next minor release?

robjtede:
Reviewing this was very helpful in gaining a better understanding of how the macros work. So, I've made some simplifications that reduce noise in the diff. While giving the same API diff:

```
 Documenting async-compression v0.4.0 (/Users/rob/Development/forks/async-compression)
    Finished dev [unoptimized + debuginfo] target(s) in 0.48s
    Blocking waiting for file lock on package cache
 Documenting async-compression v0.4.0 (/Users/rob/Development/forks/async-compression)
    Finished dev [unoptimized + debuginfo] target(s) in 0.53s
Removed items from the public API
=================================
(none)

Changed items in the public API
===============================
(none)

Added items to the public API
=============================
+pub fn async_compression::futures::bufread::ZstdDecoder<R>::with_dict(inner: R, dictionary: &[u8]) -> std::io::error::Result<Self>
+pub fn async_compression::futures::bufread::ZstdEncoder<R>::with_dict(inner: R, level: async_compression::Level, dictionary: &[u8]) -> std::io::error::Result<Self>
+pub fn async_compression::futures::write::ZstdDecoder<W>::with_dict(inner: W, dictionary: &[u8]) -> std::io::error::Result<Self>
+pub fn async_compression::futures::write::ZstdEncoder<W>::with_dict(inner: W, level: async_compression::Level, dictionary: &[u8]) -> std::io::error::Result<Self>
+pub fn async_compression::tokio::bufread::ZstdDecoder<R>::with_dict(inner: R, dictionary: &[u8]) -> std::io::error::Result<Self>
+pub fn async_compression::tokio::bufread::ZstdEncoder<R>::with_dict(inner: R, level: async_compression::Level, dictionary: &[u8]) -> std::io::error::Result<Self>
+pub fn async_compression::tokio::write::ZstdDecoder<W>::with_dict(inner: W, dictionary: &[u8]) -> std::io::error::Result<Self>
+pub fn async_compression::tokio::write::ZstdEncoder<W>::with_dict(inner: W, level: async_compression::Level, dictionary: &[u8]) -> std::io::error::Result<Self>


```

robjtede:
Note on simplification:

The actual problem to be solved was adding custom constructors to the decoders _as well_ as encoders. So instead of poluting the type parameters at callsites, I've just added a way in the macro to do so, with clear markers.

Future work: the runtime encoder/decoder macros know what their generics should be. It it probbaly possible to omit the $inner param from the algos! macro.

robjtede (approved):

robjtede:
Added some testing for this feature.

@NobodyXu if you're happy with my changes we can merge this.

robjtede:
Oh great... Windows-only test failures.

miam-miam100:
Thanks @robjtede, sorry I couldn't help I've been competing in the GMTK game jam

tests/artifacts/lib.rs
```diff
@@ -0,0 +1,352 @@
+//! Adaptors between compression crates and Rust's modern asynchronous IO types.
+//!
```
NobodyXu, line 2: What is this file for?

NobodyXu (changes requested):

tests/artifacts/lib.rs
robjtede: it _is_ the test case

NobodyXu: But this file `tests/artifacts/lib.rs` contains duplicate code from [`src/lib.rs`](https://github.com/Nullus157/async-compression/blob/main/src/lib.rs).

Did you accidentally copied `src/lib.rs` to `testrs/artifacts/lib.rs`?

NobodyXu: Oh I see, it's the file being compressed.
I thought it was some duplicate code in tests and it got me confused.

NobodyXu (approved):

NobodyXu:
P.S. @robjtede I think using squash merge is better than merge commits, especially in this PR where I merge changes in main back to this PR.

With squash merge, the commit history would be much more straight-forward.

If you agree, I will disable merge commit in setting and only allow squash merging.

NobodyXu:
I've disabled merge commit and rebase merge, allowing only squash merge.

PR #238, opened 2023-07-29
feat: deflate64 decompression (decode-only) support

### LINKED ISSUE #237: deflate64 support
Windows may creates zip file with deflate64 compression method.
I want to decompress such a zip file with deflate64 so I want async-compression to support deflate64.

anatawa12:
Closes #237

Because I could not find deflate64 compression implementation, this only adds encode support for deflate64

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ pin-project-lite = "0.2"
 tokio = { version = "1.24.2", optional = true, default-features = false }
 xz2 = { version = "0.1.6", optional = true }
 zstd-safe = { version = "6", optional = true, default-features = false }
-deflate64 = { version = "0.1.2", optional = true }
+deflate64 = { version = "0.1.5", optional = true }
 
 [dev-dependencies]
 bytes = "1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 # groups
 all = ["all-implementations", "all-algorithms"]
 all-implementations = ["futures-io", "tokio"]
-all-algorithms = ["brotli", "bzip2", "deflate", "gzip", "lzma", "xz", "zlib", "zstd"]
+all-algorithms = ["brotli", "bzip2", "deflate", "gzip", "lzma", "xz", "zlib", "zstd", "deflate64"]
 
 # algorithms
 deflate = ["flate2"]
@@ -29,6 +29,7 @@ xz = ["xz2"]
 zlib = ["flate2"]
 zstd = ["libzstd", "zstd-safe"]
 zstdmt = ["zstd", "zstd-safe/zstdmt"]
+deflate64 = ["dep:deflate64"]
 
 [dependencies]
 brotli = { version = "3.3", optional = true, default-features = false, features = ["std"] }
@@ -42,6 +43,7 @@ pin-project-lite = "0.2"
 tokio = { version = "1.24.2", optional = true, default-features = false }
 xz2 = { version = "0.1.6", optional = true }
 zstd-safe = { version = "6", optional = true, default-features = false }
+deflate64 = { version = "0.1.2", optional = true }
 
 [dev-dependencies]
 bytes = "1"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0), 
 
 ## Unreleased
 
+- Add `Deflate64Decoder` behind new crate feature `deflate64`.
+
 ## 0.4.1 - 2023-07-10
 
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -42,6 +43,7 @@ pin-project-lite = "0.2"
 tokio = { version = "1.24.2", optional = true, default-features = false }
 xz2 = { version = "0.1.6", optional = true }
 zstd-safe = { version = "6", optional = true, default-features = false }
+deflate64 = { version = "0.1.2", optional = true }
```
NobodyXu, line 46: This dependency looks quite new and its doc is a bit lacking.

I'm concern about if it will get enough maintanence.

NobodyXu (changes requested):

Cargo.toml
anatawa12: That's my crate and I'll maintenance as possible. I'm going to add more docs.

NobodyXu: Yes I noticed that, but I am still concerned about the bus factor.

I have two suggestions on this:
 - Invite people to your repository and its crates.io package to reduce bus factor
 - Optionally, setup automatic release if you don't want to invite them to the crates.io package
 - Give people you trusted with admin access

Alternatively, you can consider making this a rust-lang projects and that will make everybody on rust-lang able to maintain it.

robjtede: I have to agree with @NobodyXu here since this crate is widely used. I'm tracking the discussion on https://github.com/zip-rs/zip about deflate64 support; if they had it we'd probably prefer to use it.

I'd also accept this PR given co-ownership of `deflate64` on crates.io.

NobodyXu: @robjtede We should also do a review of `deflate64` and it will be great if we can have a few protests to assert that anything compressed by zip on windows can be decompressed by `deflate64` to original crates.

Unfortunately I'm not very familiar with deflate64 and other algorithms, so I would have a look at the code but I would definitely want someone who also know deflate64 to be the maintainer of `deflate64`.

robjtede: > We should also do a review of deflate64

> definitely want someone who also know deflate64 to be the maintainer

For sure; and I'm not in any rush to merge this so IMO lets see if zip-rs makes any movement in the next week.

anatawa12: > Invite people to your repository and its crates.io package to reduce bus factor

Sorry, I don't have friends who write rust so I could not invite friends to my repository.

> Alternatively, you can consider making this a rust-lang projects and that will make everybody on rust-lang able to maintain it.

I don't know how to make an project a rust-lang project but If there is one way to do that, I think this is better since I have no friends. Could you let me know how to do this?

NobodyXu: > Sorry, I don't have friends who write rust so I could not invite friends to my repository.

You can invite @robjtede or me.

> I don't know how to make an project a rust-lang project but If there is one way to do that, I think this is better since I have no friends. Could you let me know how to do this?

I'm not sure either, but you can ask on zulip for this and you will get answers quickly.

robjtede:
```bash
cargo owner --add <username>
# or
cargo owner --add github:<team>
```

NobodyXu: Thanks @anatawa12 I've accepted the invitation.

robjtede:
Using a modified example, I couldn't get [the test file](https://github.com/dotnet/runtime-assets/blob/95277f38e68b66f1b48600d90d456c32c9ae0fa2/src/System.IO.Compression.TestData/ZipTestData/compat/deflate64.zip) to decompress.

```
 › cargo run --example zstd_gzip --features="tokio,zstd,gzip,deflate64" < examples/deflate64.zip
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `/Users/rob/.cargo/target-shared/debug/examples/zstd_gzip`
Error: Custom { kind: InvalidData, error: "invalid data" }
```

anatawa12:
The `deflate64.zip` file in dotnet assets is an ZIP file with entry compressed with deflate64, not a valid defalte64 stream.
So, passing the zip file itself to the decoder will cause an error.

On the test in my crate, I manually parsed the zip file header with binary editor and I'm using their knowledge of offset and length of deflate64 stream in the zip file.

https://github.com/anatawa12/deflate64-rs/blob/64349cd3982ce8b88117447c4036853a5e07e64d/tests/stream.rs#L4-L5

robjtede:
Ahh I see, well I'm just trying to validate the implementation before approving. Can you suggest something?

anatawa12:
As I did before, you can extract deflate64 stream from zip file manually.
As far as I know, there are no raw deflate64 compressor tool so I think this is only way for  now.

robjtede (approved):

robjtede:
Verified the functionality as you showed in the deflate64 repo.

NobodyXu (approved):

Issue #85, opened 2020-04-19
Test minimal versions


Nemo157:
So many blockers though 😭

PR #249, opened 2023-10-04
Fix a minimum-versions issue

g2p:
Tested with:

    cargo +nightly update -Zminimal-versions; cargo check --all-features

See https://github.com/rust-lang/flate2-rs/pull/221
(And #85 would be nice)

### DIFF
--- a/.github/workflows/exhaustive.yml
+++ b/.github/workflows/exhaustive.yml
@@ -26,7 +26,7 @@ jobs:
     - run: cargo --locked nextest run --workspace --all-features
     - run: cargo --locked test --workspace --doc --all-features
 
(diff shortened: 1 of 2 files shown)


NobodyXu (changes requested):
Can you add a new job to CI which tests the minimal versions?

g2p:
> Can you add a new job to CI which tests the minimal versions?

Added!

NobodyXu (approved):

NobodyXu:
Thank you @g2p

robjtede:
It's nice that the tools to catch this sort of stuff are getting easier to use!

PR #252, opened 2023-11-14
expose memlimit settings for xz2 decoders

svenrademakers:
xz2 decoders are by default configured with a memory limit value of u64::max. This is problematic on targets where there is limited memory available for decoding. e.g. one could run out of memory. By exposing an additional function call `XzDecoder::with_memlimit(..)` and `LzmaDecoder::with_memlimit(..)` users can initialize a decoder with a given memory limit.

Consequently specifying a too low value means the decoder will fail to decompress. On this occasion the decoder will return an `io::Error::Custom` value with the message "MemLimit" to signal it needs to be configured with more memory.

This error is propagated by the chosen interface, `AsyncBufRead` or `AsyncWrite`, therefore no additional error handling has to be implemented to handle this "side effect"

### DIFF
--- a/src/codec/lzma/decoder.rs
+++ b/src/codec/lzma/decoder.rs
@@ -10,7 +10,13 @@ pub struct LzmaDecoder {
 impl LzmaDecoder {
     pub fn new() -> Self {
         Self {
-            inner: crate::codec::Xz2Decoder::new(),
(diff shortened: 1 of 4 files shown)


NobodyXu (approved):
LGTM

robjtede (approved):

PR #254, opened 2023-12-02
Exhaust AsyncBufRead reader

svenrademakers:
When the decoder goes into State::Done, the decoder is considered done with reading. Therefore no calls will be made to the underlying reader. As a consequence, the reader never reaches EOF state. i.e. The AsyncBufRead never returns `Poll::Ready(Ok(&[]))`. This is problematic as this prevents underlaying streams, readers and channels to gracefully shutdown.

To solve this issue, the decoder will make one interation more into state::Decoding. This forces one last read into the underlying reader by calling poll_fill_buffer. before it goes into State::Done via State::Flushing.

### DIFF
--- a/src/futures/bufread/generic/decoder.rs
+++ b/src/futures/bufread/generic/decoder.rs
@@ -79,11 +79,7 @@ impl<R: AsyncBufRead, D: Decode> Decoder<R, D> {
                         let done = this.decoder.decode(&mut input, output)?;
                         let len = input.written().len();
                         this.reader.as_mut().consume(len);
-                        if done {
-                            State::Flushing
-                        } else {
-                            State::Decoding
-                        }
+                        State::Decoding
                     }
                 }
 
--- a/src/tokio/bufread/generic/decoder.rs
+++ b/src/tokio/bufread/generic/decoder.rs
@@ -79,11 +79,7 @@ impl<R: AsyncBufRead, D: Decode> Decoder<R, D> {
                         let done = this.decoder.decode(&mut input, output)?;
                         let len = input.written().len();
                         this.reader.as_mut().consume(len);
-                        if done {
-                            State::Flushing
-                        } else {
-                            State::Decoding
-                        }
+                        State::Decoding
                     }
                 }
 


NobodyXu:
@svenrademakers  The CI has failed due to infinite loop (just in case you didn't notice).

Honestly, I'm not sure how your PR fixed the issue your describe and also unsure about the bug you are describing.

svenrademakers:
Long story short, it's not up to the decoder to decide when the EOF state of the `AsyncRead` implementation is reached. It should be signaled by the source instead (the reader object inside the decoder). The end of data, in our case, exposed as an `AsyncBufRead`, is reached when `.poll_fill_buffer()` returns either `Poll::Ready(Err(e))` or `Poll::Ready(Ok(buffer)) if buffer.is_empty()`.

A simple example: I have a file that exists out of 2 parts. The first half contains .xz encoded binary data, and the second half contains similar binary data but is uncompressed. 

```rust
let file = OpenOptions::new().read(true).open("/tmp/my_container.img").await?;
let file_with_decoder = XzDecoder::with_mem_limit(reader, mem_limit);
tokio::io::copy(file_with_decoder, embedded_system).await?;
```

With the current codebase, only the first part of the image is copied. Now, for this given example, a fix is relatively easy to implement, but understand that this gets very awkward for more complex byte streams.

A second example: a downloaded file is checksum-calculated while being decompressed and copied. The sender needs to be exhausted so that the final checksum can be verified:

```rust 
let bytes_stream = reqwest::get(url).await?;
// Sha256StreamValidator validates when EOF state is reached and returns an Error when expected_sha != calculated sha
let reader = tokio_util::io::StreamReader::new(Sha256StreamValidator::new(reader, expected_sha));
let reader_with_decoder = XzDecoder::with_mem_limit(reader, mem_limit);
tokio::io::copy(reader_with_decoder, sink()).await?;
```

Checksum validation never happens here as the XzDecoder starves the underlying `reader,` one read call before EOF state.

From a functional perspective, the decoder is not doing anything wrong, it returns `Poll::Ready(Ok(()))` when its done decoding. But understand there is a subtle difference between the decoder being done and the reader being exhausted.

In Rust, you should be able to pull bytes out of Streams and Readers if there are still bytes available. If you have a none exhausting reader, such as a `TcpStream`, its up to the user to limit the amount of bytes that can be read, with for instance `take()`:

``` rust 
 let reader = tokio::io::repeat(123).take(1234);
let reader_with_decoder = XzDecoder::with_mem_limit(reader, mem_limit);
let copied = tokio::io::copy(&mut reader, &mut sink()).await?;
```

I agree that my proposed code changes are very naive, i will make a better implementation if we can reach a consensus on the matter i tried to explain above

Nemo157:
To explain why the design as it is, and why those tests are failing: the intent is to support cases like you say where there's compressed data followed by other data, this is done by using the fact that all the compression codecs are self-terminating so we can stop reading once the compressed data has been consumed and leave the rest of the reader available to other code that reads the following data.

If you look at [one of the failing tests](https://github.com/Nullus157/async-compression/blob/4c6ae38e14b45a40a97c566922c8065799bf57e0/tests/utils/test_cases.rs#L179):

```rust
let mut compressed = sync::compress(&[1, 2, 3, 4, 5, 6]);

compressed.extend_from_slice(&[7, 8, 9, 10]);

let input = InputStream::new(vec![compressed]);
let mut reader = bufread::from(&input);
let output = bufread::decompress(&mut reader);
let trailer = read::to_vec(reader);

assert_eq!(output, one_to_six());
assert_eq!(trailer, &[7, 8, 9, 10][..]);
```

It creates one `impl AsyncBufRead` and passes that by-ref into a decompressor, reads till the decompressor indicates it's read all the compressed data, then reads from the reader directly for the trailing data.

If you know that your `impl AsyncBufRead` contains only compressed data, you could use `.multiple_members(true)`, which once it finishes reading one series of compressed data checks whether there is another header available or if it's at EOF.

svenrademakers:
Eureka, i wrongly skipped over multi_member as i assumed this should be enabled on default. I can live with explicitly turning it on. Thanks for your explanation, it helped a lot in understanding the design.

PR #270, opened 2024-04-26
CI: Add `publish.yml`

NobodyXu:
This would make publishing new crate easier, without having to use commandline.

Just create a release on GitHub and it would automatically publishes to the crates.io

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,15 @@
+on:
+  push:
+    tags:        
+      - '*'
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


NobodyXu:
@robjtede If you are good with this PR, then I will add the crates.io token for publishing new versions of this crate to secrets, and then merge it.

robjtede:
We're on the right track here. I've used release-plz for some other things and I think that we should use that instead. It's a bit more of a robust approach.

NobodyXu:
I happen to need release-plz in my other project, so if I successfully implement it, I would also adopt it here.

NobodyXu:
Ok I figured it out, it's pretty simple though my other project can't use it for now.

Would open another PR soon

PR #278, opened 2024-05-28
Expose `total_in`/`total_out` from underlying `flate2` encoder types.

tobz:
👋🏻 

This PR is about exposing some more of the underlying `flate2` bits, as the title indicates.

Naturally, when asynchronously writing the compressed output, it can be difficult, otherwise, to determine how much data has been written so far. By exposing the helper methods like `total_in` and `total_out`, it becomes that much easier.

Some thoughts/caveats:

- I do realize that this is somewhat coupling the API to `flate2`, although with `flate2` being 1.x, it should be OK from a stability standpoint.
- Not all compression algorithms used come from `flate2`, so not all encoders can consistently expose these same methods.

### DIFF
--- a/src/codec/flate/encoder.rs
+++ b/src/codec/flate/encoder.rs
@@ -17,6 +17,10 @@ impl FlateEncoder {
         }
     }
 
+    pub(crate) fn get_ref(&self) -> &Compress {
+        &self.compress
+    }
+
     fn encode(
         &mut self,
         input: &mut PartialBuffer<impl AsRef<[u8]>>,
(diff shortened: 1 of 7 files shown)


robjtede:
Exposing a useful `-> u64` getter doesn't feel like it's exposing internals too much. I'd be comfortable committing to this API. Will wait for consensus before merging though.

robjtede (approved):

NobodyXu (approved):

tobz:
Oh wow, awesome!

Thanks so much y'all.

PR #285, opened 2024-07-14
Enable customizing Zstd decoding parameters.

### LINKED ISSUE #284: Support `new_with_params` for ZstdDecoder
`ZstdEncoder` has [with_quality_and_params](https://docs.rs/async-compression/latest/async_compression/futures/bufread/struct.ZstdEncoder.html#method.with_quality_and_params), which is useful for tweaking parameters for compression.

When attempting to decompress a Zstd-encoded file with a non-default window size, I receive the following error:

```
Frame requires too much memory for decoding
```

Since we do not expose the ability to override the default decoding parameters, there is no way to override the default frame size and successfully decompress this file.

quantatic:
closes #284 

This PR updates the minimum `zstd` version to `0.13.1`, in order to make `DParameter` `Clone+Copy`.


src/codec/zstd/decoder.rs
```diff
@@ -16,6 +16,16 @@ impl ZstdDecoder {
         }
     }
 
+    pub(crate) fn new_with_params(params: &[crate::zstd::DParameter]) -> Self {
```
NobodyXu, line 19: I think we should propagate the error from zstd?

quantatic: I'm not necessarily opposed to this, though in the interest of consistency, it appears that this library chooses to panic on construction errors for other encoders are decoders:

- https://github.com/Nullus157/async-compression/blob/7c2d5304b418254e13e810329ee0004a6484a7bd/src/codec/zstd/encoder.rs#L18
- https://github.com/Nullus157/async-compression/blob/7c2d5304b418254e13e810329ee0004a6484a7bd/src/codec/xz2/encoder.rs#L23

quantatic:
This change is technically semver-compatible with only a minor version bump on next release, though it's unclear to me how this would be compatible with a crate that locks `zstd = "=0.13.0`. Is such a change still considered semver-compatible?

src/codec/zstd/decoder.rs
NobodyXu: That's unfortunate, maybe we should add a `try_` variant for these APIs?

cc @robjtede what's your thoughts on this?

NobodyXu:
> This change is technically semver-compatible with only a minor version bump on next release, though it's unclear to me how this would be compatible with a crate that locks `zstd = "=0.13.0`. Is such a change still considered semver-compatible?

Given that we don't expose any implementation detail, I think it can be said as semver-compatible API-wise

src/codec/zstd/decoder.rs
robjtede: I think typically panics are acceptable when it comes to "programmer error" and I really, really don't expect decoding parameters to be user-input, or anything other than hardcoded. Therefore it's fine to me that this can panic. It's a fairly unrecoverable error for **both** these unwraps.

robjtede:
> Is such a change still considered semver-compatible

yes

robjtede (approved):

NobodyXu (approved):

PR #286, opened 2024-07-21
Use explicit paths rather than glob imports for zstd wrapper types.

quantatic:
#285 adds wrapper types for `DParameter`, but uses glob imports to pull the `libzstd::stream::raw::{C,D}Parameter` types into scope. This introduces a _possible_ forwards-compatability hazard if `libzstd` were to introduce `{C,D}Parameter` enum variants with the same name.

This PR uses explicit paths for referencing these wrapper types, to avoid this possibility.

### DIFF
--- a/src/zstd.rs
+++ b/src/zstd.rs
@@ -1,8 +1,5 @@
 //! This module contains zstd-specific types for async-compression.
 
-use libzstd::stream::raw::CParameter::*;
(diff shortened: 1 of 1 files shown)


robjtede (approved):

robjtede (commented):
thanks 👍🏻

Issue #123, opened 2021-06-03
Decoder might block if there is data remaining to be flushed from the codec and the stream blocks


Nemo157:
See https://github.com/gyscos/zstd-rs/issues/119, I'm pretty sure the local `bufread::Decoder` implementations suffer from this same issue. My first attempt at writing a test case failed to fail though.

nl5887:
I encountered this issue when keeping an open connection, while not sending data. This caused the remaining data to be kept in the decoder. I have fixed the issue, but will try to write a test case.

fr3akX:
could you, please, review and merge this PR?

PR #294, opened 2024-10-08
fix(tokio): attempt to decode from internal state even if nothing was read

link2xt:
Unlike zlib, miniz_oxide consumes data into internal state
even if there is not enough space in the output buffer.
Next time poll_fill_buf() is called we should try
to decode from internal state even if no new compressed data
was read.

This change is a port of fix <https://github.com/Nullus157/async-compression/issues/123> (commit https://github.com/Nullus157/async-compression/commit/22ed0ac4caf5d60628ff8836dd1dc2d3289cf43f) from `futures` to `tokio`.


NobodyXu (approved):

NobodyXu:
Thanks, that looks good to me

PR #299, opened 2024-10-12
Fix panic when decoding

NobodyXu:
Fix #298

PR #301, opened 2024-10-13
Revert "Fix panic when decoding"

NobodyXu:
Reverts Nullus157/async-compression#299


NobodyXu:
Took another look , I think the original impl was alright, it might be a decoder impl problem.

robjtede (approved):

PR #303, opened 2024-10-15
Fix panic: cannot consume from pending buffer

### LINKED ISSUE #298: panic: cannot consume from pending buffer
In https://github.com/rust-lang/crates.io/pull/9622 we upgraded crates.io to the latest `async-compression` release (v0.4.14), which appears to be causing issues on our staging environment:

https://rust-lang.sentry.io/share/issue/9e273e28600f48a5aec2ad9e127afbbf/

The code inside the `object_store` crate is throwing a "cannot consume from pending buffer" panic, but the root cause seems to be originating from the `async-compression` crate. We use both crates roughly like this:

```rust
let reader = object_store::buffered::BufReader::new(store, &meta);
let decompressor = Decompressor::from_extension(reader, path.extension())?;
```

I assume that https://github.com/Nullus157/async-compression/pull/294 is causing this issue, but I don't know the code well enough to find the exact bug in the code changes that are causing this behavior.

NobodyXu:
Fixed #298

This PR changes the decoder `do_poll_read` impl, to not advance the buffer on first flush.

The panic is because the decoder try to advance the buffer before polling the underlying buf reader.

~~In this PR I tried a different fix, by making sure `buf.consume` is always called, even on error.
I suspect that previously we didn't consume the buffer on error, and that might have caused the same data to be decompressed again.~~

~~I can't think of anywhere else that could fail, the decoder implementation looks alright.~~

### DIFF
--- a/src/futures/bufread/generic/decoder.rs
+++ b/src/futures/bufread/generic/decoder.rs
@@ -88,7 +88,7 @@ impl<R: AsyncBufRead, D: Decode> Decoder<R, D> {
                         State::Flushing
                     } else {
                         let mut input = PartialBuffer::new(input);
-                        let done = this.decoder.decode(&mut input, output).or_else(|err| {
+                        let res = this.decoder.decode(&mut input, output).or_else(|err| {
                             // ignore the first error, occurs when input is empty
                             // but we need to run decode to flush
                             if first {
@@ -96,13 +96,14 @@ impl<R: AsyncBufRead, D: Decode> Decoder<R, D> {
                             } else {
                                 Err(err)
                             }
-                        })?;
+                        });
 
                         first = false;
 
                         let len = input.written().len();
                         this.reader.as_mut().consume(len);
-                        if done {
+
+                        if res? {
                             State::Flushing
                         } else {
                             State::Decoding
--- a/src/futures/bufread/generic/decoder.rs
+++ b/src/futures/bufread/generic/decoder.rs
@@ -98,10 +98,12 @@ impl<R: AsyncBufRead, D: Decode> Decoder<R, D> {
                             }
                         });
 
-                        first = false;
+                        if !first {
+                            let len = input.written().len();
+                            this.reader.as_mut().consume(len);
+                        }
 
-                        let len = input.written().len();
-                        this.reader.as_mut().consume(len);
+                        first = false;
 
                         if res? {
                             State::Flushing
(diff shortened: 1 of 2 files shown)


NobodyXu:
@Turbo87 Can you try this PR please?

Turbo87:
yep, I'll give it a try.

Turbo87:
Bildschirmfoto 2024-10-16 um 09 32 57

unfortunately still failing 😢

NobodyXu:
Thanks.

Is the software open-source?

Can I have a look at the code and the test?

Turbo87:
> Can I have a look at the code and the test?

[the code yes](https://github.com/rust-lang/crates.io/blob/873bc1f5f87fc49ba74e5f962c2b48eba0d0e6d8/src/worker/jobs/downloads/process_log.rs#L157-L161), the test no. unfortunately we don't have a test that reproduces it. I can only run it on our staging environment where I can reproduce it. our test suite runs with an in-memory `object_store` instance instead of the S3 implementation we use on staging and production.

NobodyXu:
I've found the cause of the panic.

It is because the decoder try to advance the buffer before polling the underlying buf reader.

cc @Turbo87 I've updated the PR, can you try again please?

Thank you!

Turbo87:
the panic appears to be gone, but we're now seeing an "interval out of range" error result without a stacktrace. I will have to improve our logging a bit to figure out where exactly that is coming from.

NobodyXu:
cc @robjtede Shall we merge and publish this for now, since it at least fixes the panic for @Turbo87

Turbo87:
> we're now seeing an "interval out of range" error

it turns out that this was a bug on our side, related to how we calculate exponential backoff for failed jobs.

I can confirm that https://github.com/Nullus157/async-compression/pull/303 appears to fix the issue for us! 🎉 

thanks again! :)

NobodyXu:
Thank you!

NobodyXu:
Thank you!

NobodyXu:
cc @robjtede let's get this merged and cut a new release, as it is confirmed to fix the panic

NobodyXu:
I will get this merged and ask for review in the release PR.

robjtede:
Ahh wonderful, yes, lets get this out today.

Turbo87:
thanks again for the investigation, fix and release! I just merged the latest update into crates.io :)

Issue #246, opened 2023-09-14
`ZstdEncoder` should not panic if `tokio::io::AsyncWrite::flush` is called after `tokio::io::AsyncWrite::shutdown`


aatifsyed:
I'm hitting a `Flush after shutdown` panic, ultimately here:
https://github.com/Nullus157/async-compression/blob/ece5584ce59a683c38e94f9957969e1bdb08b665/src/tokio/write/generic/encoder.rs#L97

# Repro
Here's an minimal example, which reliably panics on my machine.
```rust
use async_compression::tokio::write::ZstdEncoder; // 0.4.3
use bytes::Bytes; // 1.5.0
use futures::{stream, StreamExt as _}; // 0.3.28
use std::io; // 1.72.0
use tokio::fs::File; // 1.32.0
use tokio_util::codec::{BytesCodec, FramedWrite}; // 0.7.8

async fn _main() -> io::Result<()> {
    let file = File::create("/dev/null").await?;
    let zstd_encoder = ZstdEncoder::new(file);
    let bytes_sink = FramedWrite::new(zstd_encoder, BytesCodec::new());
    stream::empty::<io::Result<Bytes>>()
        .forward(bytes_sink)
        .await
}

#[tokio::main]
async fn main() -> io::Result<()> {
    _main().await
}
```

# Explanation
Here's the sequence of events:
```
  ┌───────────────────────────────────┐ ┌──────────────────────────────────┬────────────────────────────────┐ 
  │ file::poll_flush -> Ready(Ok(())) │ │ file::poll_write -> Ready(Ok(9)) │ file::poll_shutdown -> Pending │ 
 ┌┴───────────────────────────────────┴┬┴──────────────────────────────────┴────────────────────────────────┴┐ ┌─────────────BOOM
 │  zstd::poll_flush -> Ready(Ok(()))  │                    zstd::poll_shutdown -> Pending                   │ │ zstd::poll_flush
┌┴─────────────────────────────────────┴─────────────────────────────────────────────────────────────────────┴┬┴─────────────BOOM
│                                      sink::poll_close -> Pending                                            │  sink::poll_close
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────┴──────────────BOOM
```
- The sink needs to [`close`](https://docs.rs/futures/0.3.28/futures/sink/trait.Sink.html#tymethod.poll_close)
[...]

PR #309, opened 2024-10-31
fix: Encoder state machine tolerates being wrapped by an `AsyncWrite`

### LINKED ISSUE #308: `Encoder` does not follow the `tokio::io::AsyncWrite` protocol
Until Encoder returns Poll::Ready from poll_shutdown, it should expect calls to poll_flush and poll_shutdown any number of times, and not return an error.

This is the same as https://github.com/Nullus157/async-compression/issues/246

Test case available at #309

aatifsyed:
Simplified the `Encoder` state machine.
It should now handle `poll_flush` calls after `poll_shutdown` calls.

Closes #308
See also #246

### DIFF
--- a/src/tokio/write/generic/encoder.rs
+++ b/src/tokio/write/generic/encoder.rs
@@ -100,18 +100,12 @@ impl<W: AsyncWrite, E: Encode> AsyncWrite for Encoder<W, E> {
 
     fn poll_shutdown(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
         let mut this = self.project();
-        if !*this.finished {
-            loop {
-                let mut space =
-                    PartialBuffer::new(ready!(this.writer.as_mut().poll_partial_flush_buf(cx))?);
-                let finished = this.encoder.finish(&mut space)?;
-                let bytes_encoded = space.written().len();
-                this.writer.as_mut().produce(bytes_encoded);
-                if finished {
-                    *this.finished = true;
-                    break;
-                }
-            }
+        while !*this.finished {
+            let mut space =
+                PartialBuffer::new(ready!(this.writer.as_mut().poll_partial_flush_buf(cx))?);
+            *this.finished = this.encoder.finish(&mut space)?;
+            let bytes_encoded = space.written().len();
+            this.writer.as_mut().produce(bytes_encoded);
         }
         this.writer.poll_shutdown(cx)
     }


NobodyXu:
Thanks!

You are right, poll_shutdown should indeed call self.poll_flush before calling inner.poll_shutdown

NobodyXu (changes requested):

src/tokio/write/generic/encoder.rs
```diff
@@ -13,20 +13,13 @@ use futures_core::ready;
...
-}
-
 pin_project! {
     #[derive(Debug)]
     pub struct Encoder<W, E> {
         #[pin]
         writer: BufWriter<W>,
         encoder: E,
-        state: State,
+        finished: bool
```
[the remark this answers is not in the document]
NobodyXu: In that case, can we remove `finished` as well?

tests/issues.rs
```diff
@@ -0,0 +1,153 @@
+#![cfg(all(feature = "tokio", feature = "zstd"))]
```
NobodyXu, line 1: How about naming this file `issue_246.rs`?

tests/issues.rs
```diff
@@ -0,0 +1,153 @@
...
+    tracing_subscriber::fmt()
+        .without_time()
+        .with_ansi(false)
+        .with_level(false)
+        .with_test_writer()
+        .with_target(false)
+        .with_span_events(FmtSpan::NEW)
+        .init();
+    let mut zstd_encoder = Wrapper::new(Trace::new(ZstdEncoder::new(DelayedShutdown::default())));
+    futures::executor::block_on(zstd_encoder.shutdown()).unwrap();
```
NobodyXu, line 37: I think we should test `poll_shutdown` by keeping track of numbers of calls to underlying `poll_flush`?

NobodyXu (commented):
Thanks, mostly LGTM

cc @robjtede can you take a look at this please?

tests/issues.rs
NobodyXu: Hmmm yeah I'm fine with that

tests/issues.rs
NobodyXu: I'd want the `poll_flush` to be tested, by making sure it is called if and only if `poll_shutdown` is called.

NobodyXu: Bit busy now, but in general, I think we want something like this:

```
struct DummyWriter {
    is_poll_flush_called: boolean,
    is_poll_shutdown_called: usize,
}

impl AsyncWrite {
    fn poll_flush(...) {
        assert!(!self.is_poll_shutdown_called);
        self.is_poll_flush_called = true;
        Ready(Ok())
    }

    fn poll_shutdown(...) {
        self.is_poll_shutdown_called = true;
        Ready(Ok())
    }
}
```

And then after `shutdown` is called, we check that both is set to `true` to verify this fix.

aatifsyed:
What's the level of enthusiasm on this PR?

NobodyXu:
> What's the level of enthusiasm on this PR?

I can continue review once the feedback has been addressed

PR #314, opened 2024-11-23
Adjust `Level::Precise` clamp range for flate2

### LINKED ISSUE #313: Gzip compression level 0
Currently, the compression level for Gzip is clamped between 1 (fast) and 9 (best). However, level 0 is supported by the flate2 crate and Python's gzip module. While changing the level range could be a breaking change, are there any alternative approaches to address this issue? For example, introducing a `none` level might be a viable solution?

eth3lbert:
Previously, the clamp range was from "fastest" to "best" (1-9). This PR changes the range to start from "none" (0) to allow for no compression.

Closes #313

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -216,6 +216,7 @@ impl Level {
     fn into_flate2(self) -> flate2::Compression {
         let fastest = flate2::Compression::fast();
         let best = flate2::Compression::best();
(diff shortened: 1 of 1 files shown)


NobodyXu (approved):

NobodyXu:
Thanks LGTM

PR #316, opened 2024-11-26
Add assertion in `produce` method to check buffer capacity in implementations for `BufWriter`

pluveto:


NobodyXu (approved):
Thank you!

I think it makes sense to panic if implementation of I/O objects is wrong.

How do you think @robjtede ?

src/tokio/write/buf_writer.rs
```diff
@@ -193,7 +193,12 @@ impl<W: AsyncWrite> AsyncBufWrite for BufWriter<W> {
     }
 
     fn produce(self: Pin<&mut Self>, amt: usize) {
-        *self.project().buffered += amt;
+        let this = self.project();
+        assert!(
```
NobodyXu, line 197:
```suggestion
        debug_assert!(
```

src/futures/write/buf_writer.rs
```diff
@@ -193,7 +193,12 @@ impl<W: AsyncWrite> AsyncBufWrite for BufWriter<W> {
     }
 
     fn produce(self: Pin<&mut Self>, amt: usize) {
-        *self.project().buffered += amt;
+        let this = self.project();
+        assert!(
```
NobodyXu, line 197:
```suggestion
        debug_assert!(
```

NobodyXu (changes requested):

NobodyXu:
> I generally prefer `debug_assert!` so libraries do not panic in production. Using `assert!` is not much better than using `.unwrap()` "because this should never return an error". It is very hard to debug a mobile app and get logs from a user when the app crashes because something in compression library decided to panic.

Thanks I think that makes sense to me

NobodyXu (approved):

PR #325, opened 2025-01-31
factor out codecs and core traits

### LINKED ISSUE #324: factor out `codec` into a separate crate
I would like to use the abstractions provided for the various compression algorithms in the `codec` module in a synchronous context. Would you be open to factoring out the `codec` module into a separate crate? I would be glad to work on a corresponding PR.

koehlma:
This PR fixes #324.

It introduces two new crates:

- `compression-core`: Contains the `Decode` and `Encode` traits.
- `compression-codecs`: Contains the former `codec` module.

I ran the changes through `cargo semver-checks` and it did not complain.

Along the way, I made some decisions that may require further discussion:

- The root crate `async-compression` does not expose any of the data structures of the other crates. In particular, I did **not** move the `Level` enum and the modules `zstd` and `brotli` into `compression-codecs`. This enables future breaking changes in `compression-codecs` without breaking `async-compression`.
- The crate `compression-codecs` re-exports the underlying compression libraries. The root crate does not directly depend on any of the libraries anymore. I migrated everything, including the tests, to the re-exported libraries. It may be worthwhile to add additional abstractions to `compression-codecs` to hide the underlying libraries. However, I think, that this should not be part of this PR, as it is already rather huge.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,8 +30,6 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-default = ["all"]
-
 # groups
 all = ["all-implementations", "all-algorithms"]
 all-implementations = ["futures-io", "tokio"]


Cargo.toml
```diff
@@ -1,49 +1,64 @@
...
+repository.workspace = true
+keywords = ["compression", "gzip", "zstd", "brotli", "async"]
+categories = ["compression", "asynchronous"]
 
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
+default = ["all"]
```
NobodyXu, line 33: Hmmm why is we adding this default?

It would be too aggressive and opt in features users don't need

Cargo.toml
```diff
@@ -1,49 +1,64 @@
...
+flate2 = ["compression-codecs/flate2"]
+gzip = ["compression-codecs/gzip", "flate2"]
+lzma = ["compression-codecs/lzma", "xz2"]
+xz = ["compression-codecs/xz", "xz2"]
+xz2 = ["compression-codecs/xz2"]
+zlib = ["compression-codecs/zlib", "flate2"]
+zstd = ["compression-codecs/zstd"]
+zstdmt = ["compression-codecs/zstdmt"]
+libzstd = ["compression-codecs/libzstd"]
+zstd-safe = ["compression-codecs/zstd-safe"]
```
NobodyXu, line 54:
```suggestion
```

I don't think we should forward this two features, which is enabled when zstd is enabled

NobodyXu (changes requested):
Thank you!

Some feedback on the code

Cargo.toml
koehlma: Good catch. This was not intentional.

Cargo.toml
koehlma: Technically, not having these features is a breaking change as we currently have them. I kept them for this reason. I would suggest keeping them and removing them with the next bump of the minor version.

NobodyXu: Thanks I didn't realize that it's implicitly added due to not using `deps:`

NobodyXu: Thanks I didn't realize that it's implicitly added due to not using `deps:`

bryantbiggs:
@koehlma any plans to continue on this PR?

koehlma:
Yeah, I still think it makes sense to separate those out. I would need to have a look at the changes that happened in the meantime, as they seem to break part of this PR. There was only one failing test (prior to the conflicting changes), however, I did not understand why it was failing. It would be great if there was someone who would look into that, if that were to happen again, so that we can see this through together.

I do not quite understand the A-semver-major label: This PR is designed to not be a breaking change. At least this was the intention and I don't see where it would break semver compatibility.

NobodyXu:
> I do not quite understand the A-semver-major label

I think it might be due to new crate added.

NobodyXu:
> I do not quite understand the A-semver-major label

I think it might be due to new crate added.

NobodyXu:
Can you resolve the conflicts please, so that it would trigger another CI run?

PR #327, opened 2025-02-27
encoder: ensure that flush is finished

surban:
Once a flush has been started some encoders (for example xz2) require that the flush is completed before writing more data.


NobodyXu (approved):
Thank you for fixing the bug!

PR #331, opened 2025-03-01
feat: add lz4 support

arpankapoor:
This PR adds LZ4 support, drawing from the gzip (encoder) and bzip2 (decoder) implementations.

### DIFF
--- a/src/codec/lz4/decoder.rs
+++ b/src/codec/lz4/decoder.rs
@@ -1,0 +1,93 @@
+use std::io::Result;
+
+use lz4::liblz4::{
+    check_error, LZ4FDecompressionContext, LZ4F_createDecompressionContext, LZ4F_decompress,
+    LZ4F_freeDecompressionContext, LZ4F_resetDecompressionContext, LZ4F_VERSION,
+};
+
+use crate::{codec::Decode, unshared::Unshared, util::PartialBuffer};
+
+#[derive(Debug)]
+struct DecoderContext {
+    ctx: LZ4FDecompressionContext,
+}
+
+#[derive(Debug)]
+pub struct Lz4Decoder {
+    ctx: Unshared<DecoderContext>,
+}
+
+impl DecoderContext {
+    fn new() -> Result<Self> {
+        let mut context = LZ4FDecompressionContext(core::ptr::null_mut());
+        check_error(unsafe { LZ4F_createDecompressionContext(&mut context, LZ4F_VERSION) })?;
+        Ok(Self { ctx: context })
+    }
+}
+
+impl Drop for DecoderContext {
+    fn drop(&mut self) {
+        unsafe { LZ4F_freeDecompressionContext(self.ctx) };
+    }
+}
+
+impl Lz4Decoder {
+    pub fn new() -> Self {
+        Self {
+            ctx: Unshared::new(DecoderContext::new().unwrap()),
+        }
+    }
+}
+
+impl Decode for Lz4Decoder {
+    fn reinit(&mut self) -> Result<()> {
+        unsafe { LZ4F_resetDecompressionContext(self.ctx.get_mut().ctx) };
+        Ok(())
+    }
+
+    fn decode(
+        &mut self,
+        input: &mut PartialBuffer<impl AsRef<[u8]>>,
+        output: &mut PartialBuffer<impl AsRef<[u8]> + AsMut<[u8]>>,
+    ) -> Result<bool> {
+        let mut output_size = output.unwritten().len();
+        let mut input_size = input.unwritten().len();
+        let remaining = unsafe {
+            check_error(LZ4F_decompress(
+                self.ctx.get_mut().ctx,
+                output.unwritten_mut().as_mut_ptr(),
+                &mut output_size,
+                input.unwritten().as_ptr(),
+                &mut input_size,
+                core::ptr::null(),
+            ))?
+        };
+        input.advance(input_size);
+        output.advance(output_size);
+        Ok(remaining == 0)
+    }
+
+    fn flush(
+        &mut self,
+        output: &mut PartialBuffer<impl AsRef<[u8]> + AsMut<[u8]>>,
+    ) -> Result<bool> {
+        self.decode(&mut PartialBuffer::new(&[][..]), output)?;
+
+        loop {
+            let old_len = output.written().len();
+            self.decode(&mut PartialBuffer::new(&[][..]), output)?;
+            if output.written().len() == old_len {
+                break;
+            }
+        }
+
+        Ok(!output.unwritten().is_empty())
+    }
+
+    fn finish(
+        &mut self,
+        output: &mut PartialBuffer<impl AsRef<[u8]> + AsMut<[u8]>>,
+    ) -> Result<bool> {
+        Ok(true)
+    }
+}
--- a/src/codec/lz4/decoder.rs
+++ b/src/codec/lz4/decoder.rs
@@ -88,6 +88,6 @@ impl Decode for Lz4Decoder {
         &mut self,
         output: &mut PartialBuffer<impl AsRef<[u8]> + AsMut<[u8]>>,
     ) -> Result<bool> {
-        Ok(true)
+        self.flush(output)
     }
 }
--- a/src/codec/lz4/decoder.rs
+++ b/src/codec/lz4/decoder.rs
@@ -32,7 +32,7 @@ impl Drop for DecoderContext {
 }
 
 impl Lz4Decoder {
-    pub fn new() -> Self {
+    pub(crate) fn new() -> Self {
         Self {
             ctx: Unshared::new(DecoderContext::new().unwrap()),
         }
--- a/src/codec/lz4/encoder.rs
+++ b/src/codec/lz4/encoder.rs
@@ -7,33 +7,50 @@ use lz4::liblz4::{
     LZ4F_freeCompressionContext, LZ4F_VERSION,
 };
 
-use crate::{codec::Encode, unshared::Unshared, util::PartialBuffer};
+use crate::{codec::Encode, lz4::EncoderParams, unshared::Unshared, util::PartialBuffer};
+
+// https://github.com/lz4/lz4/blob/9d53d8bb6c4120345a0966e5d8b16d7def1f32c5/lib/lz4frame.h#L281
+const LZ4F_HEADER_SIZE_MAX: usize = 19;
 
 #[derive(Debug)]
 struct EncoderContext {
     ctx: LZ4FCompressionContext,
 }
 
-#[derive(Debug)]
+#[derive(Clone, Copy, Debug)]
 enum State {
-    Init {
-        preferences: LZ4FPreferences,
-        buffer: PartialBuffer<Vec<u8>>,
-    },
-    Encoding {
-        buffer: PartialBuffer<Vec<u8>>,
-    },
-    Footer {
-        buffer: PartialBuffer<Vec<u8>>,
-    },
+    Header,
+    Encoding,
+    Footer,
     Done,
+}
+
+enum Lz4Fn<'a, T>
+where
+    T: AsRef<[u8]>,
+{
+    Begin,
+    Update { input: &'a mut PartialBuffer<T> },
+    Flush,
+    End,
 }
 
 #[derive(Debug)]
 pub struct Lz4Encoder {
     ctx: Unshared<EncoderContext>,
     state: State,
+    preferences: LZ4FPreferences,
     limit: usize,
+    maybe_buffer: Option<PartialBuffer<Vec<u8>>>,
+    /// Minimum dst buffer size for a block
+    block_buffer_size: usize,
+    /// Minimum dst buffer size for flush/end
+    flush_buffer_size: usize,
+}
+
+// minimum size of destination buffer for compressing `src_size` bytes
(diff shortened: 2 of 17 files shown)


src/codec/lz4/encoder.rs
```diff
@@ -0,0 +1,274 @@
...
+
+impl Encode for Lz4Encoder {
+    fn encode(
+        &mut self,
+        input: &mut PartialBuffer<impl AsRef<[u8]>>,
+        output: &mut PartialBuffer<impl AsRef<[u8]> + AsMut<[u8]>>,
+    ) -> Result<()> {
+        loop {
+            match &mut self.state {
+                State::Init {
```
NobodyXu, line 110: Can we put init into the new function instead of having a separate state?

src/codec/lz4/encoder.rs
```diff
@@ -0,0 +1,274 @@
...
+                        buffer: buffer.take(),
+                    }
+                }
+
+                State::Encoding { buffer } => {
+                    output.copy_unwritten_from(buffer);
+
+                    if buffer.unwritten().is_empty() {
+                        buffer.reset();
+                        let size = std::cmp::min(input.unwritten().len(), self.limit);
```
NobodyXu, line 126:
```suggestion
                        let size = input.unwritten().len().min(self.limit);
```

NobodyXu (changes requested):

NobodyXu:
Thank you!

I have some questions for the PR

src/codec/lz4/encoder.rs
arpankapoor: We probably could. I added it as a separate state since `LZ4F_compressBegin` is fallible. Looking at the [C code](https://github.com/lz4/lz4/blob/9d53d8bb6c4120345a0966e5d8b16d7def1f32c5/lib/lz4frame.c#L690-L814), I see that the only failure it could have is an allocation failure, which should be rare I guess?

src/codec/lz4/encoder.rs
```diff
@@ -0,0 +1,274 @@
...
+                } => {
+                    Self::init(self.ctx.get_mut().ctx, preferences, buffer);
+
+                    self.state = State::Encoding {
+                        buffer: buffer.take(),
+                    }
+                }
+
+                State::Encoding { buffer } => {
+                    output.copy_unwritten_from(buffer);
```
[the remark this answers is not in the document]
arpankapoor: Yes, it is copying from the part written by `lz4-sys`, which is given by the vec length. It is unwritten from the `PartialBuffer`'s perspective.

src/codec/lz4/encoder.rs
NobodyXu: Yeah...we could have a falliable new as well if people care about it

src/codec/lz4/encoder.rs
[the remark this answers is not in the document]
NobodyXu: I read that the written part of PartialBuffer has been moved forward after compression is completed, so I don't think it is in unwritten?

It might be a good idea to write a simple test to make sure it works

src/codec/lz4/encoder.rs
arpankapoor: I prefer to keep `new` infallible and have the `Init` state as is if you don't feel strongly about it. If you do, what's your preference out of the 2 options we have:
1. `unwrap()` on the output of `LZ4F_compressBegin`.
2. make `new` infallible.

src/codec/lz4/encoder.rs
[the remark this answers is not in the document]
arpankapoor: I think you're confusing between the `input` `PartialBuffer` and the `buffer` stored in the state. A new round of compression is started when `buffer` is fully drained into the `output` (the `if` condition). `buffer` is reset (`buffer.reset()`), it's len set to the size of compressed data and **`input`** is advanced by the amount of data passed to `LZ4F_compressUpdate`. Very similar to how the sync [`lz4` crate does it](https://github.com/10XGenomics/lz4-rs/blob/287dbd063a53738917f1078192185839215d10a0/src/encoder.rs#L164-L183).

Could you give me an outline of the test you'd like for this? I've added the standard test suite similar to other compression algorithms for `lz4`, and those tests pass as confirmed by the CI.

src/codec/lz4/encoder.rs
NobodyXu: Hmmm thinking it again I think the current is quite good to have error reported instead of panic.

src/codec/lz4/encoder.rs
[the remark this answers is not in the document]
NobodyXu: Thanks I got it now

I see you already have that, I probably forgot that when doing the second review, sorry about that

src/codec/lz4/encoder.rs
```diff
@@ -0,0 +1,274 @@
...
+                                self.ctx.get_mut().ctx,
+                                buffer.unwritten_mut().as_mut_ptr(),
+                                buffer.get_mut().capacity(),
+                                input.unwritten().as_ptr(),
+                                size,
+                                core::ptr::null(),
+                            ))?;
+                            buffer.get_mut().set_len(len);
+                        }
+                        input.advance(size);
```
NobodyXu, line 138: Is it possible to write directly into `output`?

NobodyXu (changes requested):

src/codec/lz4/encoder.rs
arpankapoor: Quoting from the [lz4 docs](https://github.com/lz4/lz4/blob/9d53d8bb6c4120345a0966e5d8b16d7def1f32c5/lib/lz4frame.h#L325-L328):
> Important rule: dstCapacity MUST be large enough to ensure operation success even in worst case situations.
> This value is provided by LZ4F_compressBound().
> If this condition is not respected, LZ4F_compress() will fail (result is an errorCode).
> After an error, the state is left in a UB state, and must be re-initialized or freed.

So, we have some constraints on the size of output that can be given to `LZ4F_compressUpdate`. It is possible to have an appropriately sized `BufWriter` initially, but enforcing that during compression would be slightly involved. It could be addressed in a future PR if someone finds the excessive copying impacting performance.

NobodyXu: Perhaps we can do a simple optimization here: if `output.unwritten().len() >= buffer.unwritten().len()`, then we could use `output` directly.

With this optimization, we might also lazily allocate `buffer` to reduce memory usage

arpankapoor: I've refactored the code to minimize use of the internal buffer (which is lazily allocated now). So, I'd appreciate another review.

src/codec/lz4/encoder.rs
```diff
@@ -0,0 +1,335 @@
...
+                        let len = self.write(Lz4Fn::Flush::<&[u8]>, output)?;
+                        len == 0
+                    } else {
+                        false
+                    }
+                }
+
+                State::Footer => {
+                    if let Some(buffer) = self.maybe_buffer.as_mut() {
+                        output.copy_unwritten_from(buffer);
```
NobodyXu, line 257: Same here

src/codec/lz4/encoder.rs
```diff
@@ -0,0 +1,335 @@
...
+            }
+
+            if let State::Done = self.state {
+                return Ok(true);
+            }
+
+            if output.unwritten().is_empty() {
+                return Ok(false);
+            }
+        }
```
NobodyXu, line 333:
```suggestion
        self.flush(output)
```

NobodyXu (changes requested):
Thank you for taking my advice!

For the impl, I think there are some bugs and some rooms for improvements

src/codec/lz4/decoder.rs
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        Ok(!output.unwritten().is_empty())
+    }
+
+    fn finish(
+        &mut self,
+        output: &mut PartialBuffer<impl AsRef<[u8]> + AsMut<[u8]>>,
+    ) -> Result<bool> {
+        Ok(true)
```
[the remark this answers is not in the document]
arpankapoor: May want to do that in `bzip2` as well:

https://github.com/Nullus157/async-compression/blob/beb5c6b49b2fada3a9f2c1b6d8f702481f99a2fe/src/codec/bzip2/decoder.rs#L98-L103

src/codec/lz4/encoder.rs
```diff
@@ -0,0 +1,335 @@
...
+        loop {
+            match self.state {
+                State::Header => {
+                    self.write(Lz4Fn::Begin::<&[u8]>, output)?;
+                }
+
+                State::Encoding => {
+                    if let Some(buffer) = self.maybe_buffer.as_mut() {
+                        output.copy_unwritten_from(buffer);
+                    }
```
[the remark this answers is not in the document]
arpankapoor: I've moved the buffer drain inside `write`.

src/codec/lz4/encoder.rs
```diff
@@ -0,0 +1,335 @@
...
+        loop {
+            let done = match self.state {
+                State::Header => {
+                    self.write(Lz4Fn::Begin::<&[u8]>, output)?;
+                    false
+                }
+
+                State::Encoding => {
+                    if let Some(buffer) = self.maybe_buffer.as_mut() {
+                        output.copy_unwritten_from(buffer);
```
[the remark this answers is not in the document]
arpankapoor: Yes, but we'd drain as much as we can, return `Ok(false)` and drain the remaining bytes in the next call.

src/codec/lz4/encoder.rs
```diff
@@ -0,0 +1,335 @@
...
+        } else {
+            // SAFETY: buffer is initialized above incase of a non-direct operation
+            unsafe {
+                self.maybe_buffer
+                    .as_mut()
+                    .unwrap_unchecked()
+                    .get_mut()
+                    .set_len(len);
+            }
+        }
```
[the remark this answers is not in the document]
arpankapoor: Not the same. We're setting the length of the internal buffer which hasn't been drained into the output yet.

src/codec/lz4/encoder.rs
arpankapoor: `flush` is similar, but they differ in the function call in the `State::Encoding` case and when they're considered done.

src/codec/lz4/decoder.rs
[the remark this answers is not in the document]
NobodyXu: Thanks, I will open a PR on that later

NobodyXu (changes requested):

src/codec/lz4/encoder.rs
```diff
@@ -0,0 +1,309 @@
...
+            len
+        } else {
+            // SAFETY: buffer is initialized above incase of a non-direct operation
+            unsafe {
+                self.maybe_buffer
+                    .as_mut()
+                    .unwrap_unchecked()
+                    .get_mut()
+                    .set_len(len);
+            }
```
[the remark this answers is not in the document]
NobodyXu: Oh sorry I misunderstood it

src/codec/lz4/encoder.rs
```diff
@@ -0,0 +1,309 @@
...
+
+        let drained_after = if direct_output {
+            output.advance(len);
+            len
+        } else {
+            // SAFETY: buffer is initialized above incase of a non-direct operation
+            unsafe {
+                self.maybe_buffer
+                    .as_mut()
+                    .unwrap_unchecked()
```
NobodyXu, line 198: I'm slightly worried about this `unwrap_unchecked`, would be good if we can optimize this out the `unwrap` completely, i.e. by returning another option along side `dst_buffer`

NobodyXu, line 198: I'm slightly worried about this `unwrap_unchecked`, would be good if we can optimize this out the `unwrap` completely, i.e. by returning another option along side `dst_buffer`

NobodyXu (changes requested):

src/codec/lz4/encoder.rs
```diff
@@ -0,0 +1,301 @@
...
+            (output.unwritten_mut().as_mut_ptr(), None)
+        };
+
+        let len = match lz4_fn {
+            Lz4Fn::Begin => {
+                let len = check_error(unsafe {
+                    LZ4F_compressBegin(
+                        self.ctx.get_mut().ctx,
+                        dst_buffer,
+                        min_dst_size,
```
NobodyXu, line 143:
```suggestion
                        dst_buffer.unwritten().len(),
```

NobodyXu (commented):
Thank you!
It LGTM except for one comment

cc @robjtede wdyt

src/codec/lz4/encoder.rs
arpankapoor: `dst_buffer` is a mut pointer now, so this is not possible. `min_dst_size` is the worst case size of the destination buffers required by the [lz4 C functions](https://github.com/lz4/lz4/blob/9d53d8bb6c4120345a0966e5d8b16d7def1f32c5/lib/lz4frame.h#L306-L320), hence it is safe to use here instead of the actual destination buffer size.

NobodyXu: We could evaluate its length before the function call, then there would only be one active mutable borrow

PR #333, opened 2025-03-14
fix(flate encoder): do not mark internal state as flushed if we ran out of buffer space

link2xt:
There may be data remaining in the internal state of the encoder.

Without the fix on a second call to flush() when the caller makes space in the output buffer we will write nothing into the buffer.

### DIFF
--- a/src/codec/flate/encoder.rs
+++ b/src/codec/flate/encoder.rs
@@ -83,8 +83,9 @@ impl Encode for FlateEncoder {
             }
         }
 
-        self.flushed = true;
-        Ok(!output.unwritten().is_empty())
+        let internal_flushed = !output.unwritten().is_empty();
+        self.flushed = internal_flushed;
+        Ok(internal_flushed)
     }
 
     fn finish(


src/codec/flate/encoder.rs
```diff
@@ -83,8 +83,9 @@ impl Encode for FlateEncoder {
             }
         }
 
-        self.flushed = true;
-        Ok(!output.unwritten().is_empty())
+        let internal_flushed = !output.unwritten().is_empty();
+        self.flushed = internal_flushed;
```
NobodyXu, line 87: I think maybe it's better to check if the length of `output.unwritten()` has been changed?

There could be some data flushed to the buffer, but not enough to make it full, and maybe there could be morr data left to flush?

Not sure if this can happen, cc @robjtede wdyt?

NobodyXu (changes requested):

src/codec/flate/encoder.rs
NobodyXu: > The loop right above stops only when output.written() does not change anymore,

Thanks! I missed that when thinking about this


> If the buffer does not ever have enough space to encode anything with FlushCompress::Sync, then I don't know what happens.

I think we should return `Ok(false)` and prompt user to do another flush?

NobodyXu: Ok so I read the code again, I notice that it would only set it as flushed, when there are some bytes left unwritten in the buffer.

I think I was wrong about "what if the buffer was fully written and still left some data to flush", as the code would set flushed to false in that situation.

NobodyXu (approved):
Thank you! LGTM

r10s:
@link2xt thanks a lot for diving into that sea of code and delendencies 🤿 for finding and and for fixing the bug ❤️

PR #337, opened 2025-03-24
Expose total_in/total_out in DeflateEncoder

ardurin:
Expose the underlying `total_in` and `total_out` methods in `DeflateEncoder`, in the same fashion as #278.


NobodyXu (approved):
Thanks, LGTM!

PR #342, opened 2025-04-14
chore: add `uncheck-liblzma-version` feature

eatradish:
liblzma-rs requires a higher version of liblzma to allow dynamic linking by default (see https://github.com/Portable-Network-Archive/liblzma-rs/issues/173 for details), This feature will allow any version of liblzma on the system to use the dynamic linking

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ zlib = ["flate2"]
 zstd = ["libzstd", "zstd-safe"]
 zstdmt = ["zstd", "zstd-safe/zstdmt"]
 deflate64 = ["dep:deflate64"]
+uncheck-liblzma-version = ["liblzma/uncheck_liblzma_version"]
 
 [dependencies]
 brotli = { version = "7.0", optional = true }
@@ -45,7 +46,7 @@ lz4 = { version = "1.28.1", optional = true }
 memchr = "2"
 pin-project-lite = "0.2"
 tokio = { version = "1.24.2", optional = true, default-features = false }
-liblzma = { version = "0.4.0", optional = true }
+liblzma = { version = "0.4.1", optional = true }
 zstd-safe = { version = "7", optional = true, default-features = false }
 deflate64 = { version = "0.1.5", optional = true }
 


NobodyXu (commented):
Thank you, I wonder if adding another feature is necessary?

You could add a dependency to liblzma in your project and enable that feature directly

eatradish:
> Thank you, I wonder if adding another feature is necessary?
> 
> You could add a dependency to liblzma in your project and enable that feature directly

You do have a point, well I'm not sure if I should add it or not

eatradish:
But I have a question, if I define a version of liblzma-rs that doesn't match the version that async-compression depends on, will the feature be overwritten?

NobodyXu:
> But I have a question, if I define a version of liblzma-rs that doesn't match the version that async-compression depends on, will the feature be overwritten?

No you would keep it in sync

robjtede:
I don't think we should provide proxy features for any of the codecs, you're able to enable these yourself:

```toml
liblzma = { version = "0.4.1", features = ["uncheck_liblzma_version"] }
```

MingcongBai:
> I don't think we should provide proxy features for any of the codecs, you're able to enable these yourself, by adding this to your manifest:
> 
> ```toml
> liblzma = { version = "0.4.1", features = ["uncheck_liblzma_version"] }
> ```

The only thing though is that it causes difficulty for building Rust programs on older platforms (say, Ubuntu 22.04) - by our testing, liblzma before and after 5.6 is not perfectly compatible... To quote the original issue on the `liblzma-rs` end:

> However, with Ubuntu 22.04, [oma](https://github.com/aosc-dev/oma) built with statically linked xz/liblzma >= 5.6 would cause data errors during decompression (when xz(1) is called) - there seems to be some strange ABI incompatibilities between liblzma 5.2 and liblzma >= 5.6.

The approach you noted can also be quite tedious - say when `async-compression`'s dependency is updated a `liblzma` version that differs from what is overriden in `Cargo.toml`, then the feature specification would then be invalidated.

mokurin000:
> The approach you noted can also be quite tedious - say when `async-compression`'s dependency is updated a `liblzma` version that differs from what is overriden in `Cargo.toml`, then the feature specification would then be invalidated (please do correct me if I'm wrong).

It's not only invalidated, but the whole build process will fail fast. cargo would say:

```text
the package `liblzma-sys` links to the native library `lzma`, but it conflicts with a previous package which links to `lzma` as well:
package `liblzma-sys v0.4.3`
    ... which satisfies dependency `liblzma-sys = "^0.4.3"` (locked to 0.4.3) of package `liblzma v0.4.1`
    ... which satisfies dependency `liblzma = "^0"` (locked to 0.4.1) of package `testbuild v0.1.0 (/data/data/com.termux/files/home/testbuild)`
Only one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. Try to adjust your dependencies so that only one package uses the `links = "lzma"` value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.
```

But we could workaround things by simply not specifying version of `liblzma`.

```toml
[dependencies]
async-compression = { version = "0.4.22", features = ["lzma",] }
liblzma = { version = "0", features = [] }
```

NobodyXu:
> But we could workaround things by simply not specifying version of liblzma.


That's a new idea I didn't think of!

I think we can also have wildcard in version requirements, so

```toml
liblzma = { version = "*", features = [] }
```

Might work?

robjtede:
It's dangerous for async-compression to use * versions, and I usually advise to avoid them entirely.

Aside from the case with liblzma where multiple native libraries conflict, detecting duplicate packages generally is very easy with Cargo:

```
cargo tree -i liblzma
```

... will error if there are multiple liblzma in the dep tree.

PR #347, opened 2025-06-09
Fix clippy error in async-compression

NobodyXu:


NobodyXu:
Turns out that the branch protection isn't updated and still used the banned merge commit.

Updated and should have fixed it

PR #364, opened 2025-08-18
Allow configuring bzip2 to use bzip2 sys

DAlperin:
As of v0.6.0 of the bzip2 crate it defaults to a rust reimplementation of bzip rather than linking the c-lib. For various reasons I still require the ability to link the version, so I've exposed it as a feature in the off chance others have the same problem!


NobodyXu:
There's a better way: you could add the bzip2 dep to your hinary project and enable the feature there, as long as it's the same major-minor version the feature will be unified.

PR #378, opened 2025-08-26
Add repository to Cargo.toml package

atouchet:
A link to this crate's repository does not appear on crates.io and I'm assuming this is the reason for that.


robjtede:
good spot 👍🏻 thanks

superseded by #379

PR #381, opened 2025-08-31
rm unused dep from async-compression

NobodyXu:


crates/async-compression/Cargo.toml
```diff
@@ -32,41 +32,32 @@ all-algorithms = [
...
 # core dependencies
 futures-core.workspace = true
-memchr.workspace = true
 pin-project-lite.workspace = true
 compression-codecs.workspace = true
 compression-core.workspace = true
-# optionals deps
-brotli = { version = "8", optional = true }
-bzip2 = { version = "0.6", optional = true }
-flate2 = { version = "1.0.13", optional = true }
```
NobodyXu, line 60: technically this removes a feature, but in reality this feature does nothing and should never be used to begin with (should have marked them as `dep:` but)

PR #383, opened 2025-09-22
Flush compressed data out of encoders more often

lchsk:
Currently flush() is not getting called so compressed output is not returned to the client the encoder until input is fully read(). In my use case, flushing more often is necessary as otherwise it appears to the client the data gets 'buffered'.

There's been a previous PR here and some discussion https://github.com/Nullus157/async-compression/pull/155 but appears to be abandonded so if there's interest in this fix, this PR tries this again.
The tests are passing with this change.

### DIFF
--- a/crates/async-compression/src/futures/bufread/generic/encoder.rs
+++ b/crates/async-compression/src/futures/bufread/generic/encoder.rs
@@ -14,6 +14,7 @@ use pin_project_lite::pin_project;
 enum State {
     Encoding,
     Flushing,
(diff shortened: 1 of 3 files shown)


NobodyXu (changes requested):
Thank you!

can you also update the futures bufread encoder as well?

lchsk:
> Thank you!
> 
> can you also update the futures bufread encoder as well?

Sure, updated it just now.

PR #388, opened 2025-09-25
Flush data on BzDecoder::finish

NobodyXu:


NobodyXu:
cc @robjtede would need some help on this, I'm scratching my head to figure out why this failed on futures-io but not tokio

Issue #384, opened 2025-09-24
Deduplicate tokio/futures-io implementation?


NobodyXu:
(Continuation of #377)

The futures-io traits and tokio traits are almost identical except for different name for some methods.

And yet we have two separate impl with lots of duplicate code, having to maintain two implementations adds more burden.

It makes more sense to have a futures-io impl, possibly in a separate crate since futures-io is much faster to compile than tokio, and then simply have the tokio impl wraps its io traits in futures-io to use it.

NobodyXu:
Turns out that AsyncRead is indeed bit different between them, with tokio using a ReadBuf to support uninitialised buffer

Sometimes I just wish tokio-io is in a separate crate so that we don't have to pull in entire tokio

PR #391, opened 2025-10-02
Deduplicate `generic::bufread::Decoder` impl of tokio/futures-io

NobodyXu:
Ref #384


crates/async-compression/src/generic/bufread/decoder.rs
```diff
@@ -0,0 +1,126 @@
...
+impl Decoder {
+    pub fn multiple_members(&mut self, enabled: bool) {
+        self.multiple_members = enabled;
+    }
+
+    pub fn do_poll_read<D: Decode>(
+        &mut self,
+        cx: &mut Context<'_>,
+        output: &mut PartialBuffer<&mut [u8]>,
+        reader: &mut dyn AsyncBufRead,
```
NobodyXu, line 46: cc @robjtede I made this design choice since
 - for async I/O inlining usually doesn't help with performance, as it is usually the I/O that is bottleneck and inlining doesn't eliminate it
 - it would avoid monomophization

crates/async-compression/src/generic/bufread/decoder.rs
```diff
@@ -0,0 +1,126 @@
...
+pub(crate) trait AsyncBufRead {
+    fn poll_fill_buf(&mut self, cx: &mut Context<'_>) -> Poll<Result<&[u8]>>;
+    fn consume(&mut self, bytes: usize);
+}
+
+#[derive(Debug)]
+pub struct Decoder {
+    state: State,
+    multiple_members: bool,
+}
```
NobodyXu, line 26: cc @robjtede I decided to keep the state as simple as possible, anything else can be passed by parameter

crates/async-compression/src/generic/bufread/decoder.rs
```diff
@@ -0,0 +1,126 @@
...
+enum State {
+    Decoding,
+    Flushing,
+    Done,
+    Next,
+}
+
+pub(crate) trait AsyncBufRead {
+    fn poll_fill_buf(&mut self, cx: &mut Context<'_>) -> Poll<Result<&[u8]>>;
+    fn consume(&mut self, bytes: usize);
```
NobodyXu, line 19: Given that the reader is already pinned, using `&mut` is much simpler here for this wrapper method

NobodyXu (commented):
Some explanation on my design/API choice

crates/async-compression/src/generic/bufread/decoder.rs
NobodyXu: I would also say that even for in-memory objects or `AsyncBufRead` with (de)compression it wouldn't matter, because the AsyncBufRead::poll_fill_buf returns a &[u8], AFAIK this is something inline cannot optimize out since it involves a buffer and our de/encoder impl often involves complex stuff.

The on,y case it could be bottleneck is when poll_fill_buf returns very small slice or the (de)compression only consumes a very small part of it

NobodyXu: cc @robjtede does it look good to you?

I've think of an alternative design which does not need to pass the `reader` at all, it also removes the variable `first` within `do_poll_read`, making it truly stateless:

```rust
impl Decoder {
    fn do_poll_read(
        &mut self,
        output: &mut PartialBuffer<&mut [u8]>,
        decoder: &mut impl Decode,
        input: &mut PartialBuffer<&[u8]>,
        first: bool,
    ) -> ControlFlow<Result<NextAction>> {
        let mut first = true;
        loop {
            match self.state {
                Decode => {
                    // logic unchanged ...

                    if res? {
                        self.state = State::Flushing;
                        return ControlFlow::Continue(()); // Ask caller to call `reader.consume` and `reader.poll_fill_buf`
                    } else {
                        self.state = State::Decoding;
                    }
                }
            }
        }
    }
}
```

Caller of this funciton will then do:

```rust
let mut input = PartialBuffer(&[][]);
let mut first = true;
loop {
    match this.inner.do_poll_read(output, this.decoder, &mut input) {
        Continue(_) => {
            if !first {
                this.reader.consume(input.written.len());
            }
            first = false;
            input = PartialBuffer(ready!(this.reader.poll_fill_buf(cx))?);
        },
       Break(res) => break res,
    }
}
```

this design is much cleaner

NobodyXu: @robjtede I've updated the code with this new design, and use `macro_rules!` to further reduce duplicate code.

I quite like the current iteration, I think it'd make maintenance much, much simpler

PR #398, opened 2025-10-26
Implement new traits `DecodeV2`/`EncodeV2`

### LINKED ISSUE #385: Reduce mono of generics in codecs
We can modify `PartialBuffer` to return a non-generic view type, so that the implementation of en/decoders would be generic free, and avoid generates duplicate code.

### LINKED ISSUE #394: De/Encode: allow trait object usage
Currently the De/Encode trait cannot be used as trait object as all due to use of generics in its API.

I propose we could add `where Self: Sized`to the existing generic, and then add a non-generic, dyn-safe version of these methods.

The generic would then have default implementation that forwards to new dyn-safe version, which would need #385 as well

NobodyXu:
Fixed #394
Fixed #385

This new traits:
 - remove generics so it is trait object safe and avoid duplicate mono
 - use new `WriteBuffer<'_>` so we can add uninitialized buffer support to it in the future easily.

```rust
trait EncodeV2 {
    fn encode(
        &mut self,
        input: &mut PartialBuffer<&[u8]>,
        output: &mut WriteBuffer<'_>,
    ) -> Result<()>;

    /// Returns whether the internal buffers are flushed
    fn flush(&mut self, output: &mut WriteBuffer<'_>) -> Result<bool>;

    /// Returns whether the internal buffers are flushed and the end of the stream is written
    fn finish(&mut self, output: &mut WriteBuffer<'_>) -> Result<bool>;
}

trait DecodeV2 {
    /// Reinitializes this decoder ready to decode a new member/frame of data.
    fn reinit(&mut self) -> Result<()>;

    /// Returns whether the end of the stream has been read
    fn decode(
        &mut self,
        input: &mut PartialBuffer<&[u8]>,
        output: &mut WriteBuffer<'_>,
    ) -> Result<bool>;

    /// Returns whether the internal buffers are flushed
    fn flush(&mut self, output: &mut WriteBuffer<'_>) -> Result<bool>;

    /// Returns whether the internal buffers are flushed
    fn finish(&mut self, output: &mut WriteBuffer<'_>) -> Result<bool>;
}
```

`WriteBuffer` is a dedicated buffer management utilies, it is designed to support uninitialized operation in a future PR.

Its layout looks like:

```
|                                       buffer                                    |
| written and initialized | unwritten but initialized | unwritten and uninitialized
```

```rust
#[derive(Debug)]
pub struct WriteBuffer<'a> {
    buffer: &'a mut [u8],
    index: usize,
}

impl<'a> WriteBuffer<'a> {
    pub fn new_initialized(buffer: &'a mut [u8]) -> Self {
        Self { buffer, index: 0 }
    }

    pub fn written(&self) -> &[u8] {
        &self.buffer[..self.index]
    }

    /// Convenient method for `.writen().len()`
    pub fn written_len(&self) -> usize {
        self.index
    }

    /// Buffer has no spare space to write any data
    fn has_no_spare_space(&self) -> bool;

    /// Initialize all uninitialized, unwritten part to initialized, unwritten part
    fn initialize_unwritten(&mut self) {}

    /// Return initialized but unwritten part.
    fn unwritten_initialized_mut(&mut self) -> &mut [u8];

    /// Advance written index within initialized part.
    ///
    /// Note that try to advance into uninitialized part would panic.
    fn advance(&mut self, amount: usize);

    fn reset(&mut self);

    fn copy_unwritten_from<C: AsRef<[u8]>>(&mut self, other: &mut PartialBuffer<C>) -> usize;
}
```

### DIFF
--- a/crates/compression-core/src/util.rs
+++ b/crates/compression-core/src/util.rs
@@ -78,7 +78,7 @@ impl<B: AsRef<[u8]> + AsMut<[u8]>> From<B> for PartialBuffer<B> {
 ///
 /// # Layout
 ///
-/// ```
+/// ```text
 /// |                                       buffer                                    |
 /// | written and initialized | unwritten but initialized | unwritten and uninitialized
 /// ```


NobodyXu:
I'd merge this in today if no objections

PR #415, opened 2025-11-09
Optimize condition for removing written data in BufWriter

NobodyXu:
Do it strategically, when:
 - it can use a memcpy instead of memmove to remove it
 - when no space left for new data

### DIFF
--- a/crates/async-compression/src/generic/write/buf_writer.rs
+++ b/crates/async-compression/src/generic/write/buf_writer.rs
@@ -139,7 +139,10 @@ impl BufWriter {
         // when the flushed data is larger than or equal to half of yet-to-be-flushed data,
         // the copyback could use version of memcpy that do copies from the head of the buffer.
         // Anything smaller than that, an overlap would happen that forces use of memmove.
-        if self.written >= (self.buffered / 3) || self.written >= 512 || self.buffered == self.buf.len() {
+        if self.written >= (self.buffered / 3)
+            || self.written >= 512
+            || self.buffered == self.buf.len()
+        {
             self.remove_written();
         }
 


crates/async-compression/src/generic/write/buf_writer.rs
```diff
@@ -136,7 +136,7 @@ impl BufWriter {
     ) -> Poll<io::Result<&mut [u8]>> {
         ready!(self.partial_flush_buf(poll_write))?;
 
-        if self.written > 0 {
+        if (self.written * 2) >= self.buffered || self.buffer == self.buf.len() {
```
NobodyXu:
```suggestion
        if (self.written * 2) >= self.buffered || self.buffered == self.buf.len() {
```

crates/async-compression/src/generic/write/buf_writer.rs
```diff
@@ -136,7 +136,10 @@ impl BufWriter {
     ) -> Poll<io::Result<&mut [u8]>> {
         ready!(self.partial_flush_buf(poll_write))?;
 
-        if self.written > 0 {
+        // when the flushed data is larger than or equal to half of yet-to-be-flushed data,
+        // the copyback could use version of memcpy that do copies from the head of the buffer.
+        // Anything smaller than that, an overlap would happen that forces use of memmove.
+        if self.written >= (self.buffered / 3) || self.buffered == self.buf.len() {
```
NobodyXu:
```suggestion
        // If there's more than 512 bytes of flushed data, then memmove can move 512B at a time
        // using all registers of AVX256 if available, which should be fast enough.
        if self.written >= (self.buffered / 3) || self.written >= 512 || self.buffered == self.buf.len() {
```

crates/async-compression/src/generic/write/buf_writer.rs
```diff
@@ -136,7 +136,10 @@ impl BufWriter {
     ) -> Poll<io::Result<&mut [u8]>> {
         ready!(self.partial_flush_buf(poll_write))?;
 
-        if self.written > 0 {
+        // when the flushed data is larger than or equal to half of yet-to-be-flushed data,
+        // the copyback could use version of memcpy that do copies from the head of the buffer.
+        // Anything smaller than that, an overlap would happen that forces use of memmove.
+        if self.written >= (self.buffered / 3) || self || self.buffered == self.buf.len() {
```
NobodyXu:
```suggestion
        if self.written >= (self.buffered / 3) || self.written >= 512 || self.buffered == self.buf.len() {
```

crates/async-compression/src/generic/write/buf_writer.rs
```diff
@@ -136,7 +136,10 @@ impl BufWriter {
     ) -> Poll<io::Result<&mut [u8]>> {
         ready!(self.partial_flush_buf(poll_write))?;
 
-        if self.written > 0 {
+        // when the flushed data is larger than or equal to half of yet-to-be-flushed data,
+        // the copyback could use version of memcpy that do copies from the head of the buffer.
+        // Anything smaller than that, an overlap would happen that forces use of memmove.
+        if self.written >= (self.buffered / 3) || self.written >= 512 || self.buffered == self.buf.len() {
```
NobodyXu:
```suggestion
        if self.written >= (self.buffered / 3)
            || self.written >= 512
            || self.buffered == self.buf.len()
        {
```

PR #419, opened 2025-11-10
docs: safety comments

robjtede:
catching up with the recent work, i feel that requiring safety docs on all existing and new unsafe blocks would be prudent


robjtede:
having to prioritize other stuff for now

PR #454, opened 2026-02-10
On error the encoder/decoder now emit everything it was able to decode.

orium:
Before this we would return the error and whatever we had in our encode/decoded output buffer would be swallowed.


NobodyXu:
> @NobodyXu Is there anyone else that can review so this can land?

Well I was just being careful as last time I land a PR which turns out to be have some issues

I looked at this PR again and is very sure it is good, so I'd merge it and cut a release

PR #456, opened 2026-02-11
Fix update of bytes read in the encoder state machine.

### LINKED ISSUE #455: The encoder is never flushing
The encoder is never flushing. Bug introduced in 6c0835eb5dd26cb6176acd0455437be13fdaadac when the encoder implementation was deduplicated. I've identified the problem and will open a PR with the fix.

orium:
Previously we were not updating the number of bytes read: we were just changing the local binding of `read`. This means that read was always zero and therefore the encoder never flushed.

Fixes #455


orium:
I'll try to add a test for this. Right now I'm investigating the failing tests.

NobodyXu (approved):
Thank you, LGTM

I will ask @robjtede to also review just in case I'd make a mistake somewhere

PR #466, opened 2026-03-27
feat: Add Snappy support

Erouan50:
Hello there!

This pull request introduces support for **framed** snappy decoder and encoder. I'm not an expert in compression/decompression, but I feel the snappy one is quite different from the others. My work was inspired by the gzip decoding (for the state machine), the work done in the [snap crate][1] in the framing reader and writer, and of course the [actual spec][2]. One big difference is the whole frame must be buffered before compressing or decompressing it.

I have a few questions/notes:

* The "buffering" can be skipped if the input and/or output buffers are the right size. Should I implement that in this pull request?
* One test is failing (`snappy::tokio::bufread::decompress::trailer`), but I don't know how to fix it. If I understand the test correctly, it checks if the decoder can stop decoding with some trailing data in the buffer. But with the framing format, I don't know how to address that. Should I stop decoding when the chunk type is invalid, or should I stop and fail (current behavior)?
* Should I add more tests? Because the test data (`[1, 2, 3...]`) are not compressible in snappy, the encoder and decoder never compress or decompress anything. But is it the job of the proptests to create some random test data?

[1]: https://github.com/BurntSushi/rust-snappy
[2]: https://github.com/google/snappy/blob/main/framing_format.txt

### DIFF
--- a/crates/compression-codecs/src/snappy/decoder.rs
+++ b/crates/compression-codecs/src/snappy/decoder.rs
@@ -35,14 +35,14 @@ impl SnappyDecoder {
         let expected_sum = u32::from_le_bytes(expected_sum);
 
         self.out_buf.reset();
-        let mut out_buf = self.out_buf.get_mut();
+        let out_buf = self.out_buf.get_mut();
         out_buf.clear();
         match chunk_type {
             ChunkType::Compressed => {
                 let uncompress_length = snap::raw::decompress_len(data)?;
                 out_buf.resize(uncompress_length, 0);
                 let mut decoder = snap::raw::Decoder::new();
-                decoder.decompress(&data, &mut out_buf)?;
+                decoder.decompress(data, out_buf)?;
             }
             ChunkType::Uncompressed => out_buf.extend_from_slice(data),
             _ => unreachable!(
@@ -51,7 +51,7 @@ impl SnappyDecoder {
             ),
         };
 
-        let got_sum = crc32c_masked(&out_buf);
+        let got_sum = crc32c_masked(out_buf);
         if expected_sum != got_sum {
             return Err(io::Error::new(
                 io::ErrorKind::InvalidData,
--- a/crates/compression-codecs/src/snappy/decoder.rs
+++ b/crates/compression-codecs/src/snappy/decoder.rs
@@ -1,53 +1,67 @@
-use crate::snappy::{mask_crc, ChunkType, FrameHeader};
+use crate::snappy::{mask_crc, ChunkType, FrameHeader, MAX_BLOCK_SIZE, MAX_FRAME_SIZE};
 use crate::DecodeV2;
 use compression_core::util::{PartialBuffer, WriteBuffer};
 use std::convert::TryInto;
 use std::{io, mem};
 
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct SnappyDecoder {
     state: State,
+    in_buf: Vec<u8>,
+    out_buf: PartialBuffer<Vec<u8>>,
+}
+
+impl Default for SnappyDecoder {
+    fn default() -> Self {
+        Self {
+            state: State::default(),
+            in_buf: Vec::with_capacity(MAX_FRAME_SIZE),
+            out_buf: PartialBuffer::new(Vec::with_capacity(MAX_BLOCK_SIZE)),
+        }
+    }
 }
 
 impl SnappyDecoder {
     pub fn new() -> Self {
         Self::default()
     }
-}
 
-fn decode_chunk(chunk_type: ChunkType, mut buffer: Vec<u8>) -> std::io::Result<Vec<u8>> {
-    let data = buffer.split_off(4);
+    fn decode_chunk(&mut self, chunk_type: ChunkType) -> std::io::Result<()> {
+        let (expected_sum, data) = self.in_buf.split_at(4);
 
-    let expected_sum: [u8; 4] = buffer
-        .try_into()
-        .map_err(|_| io::Error::new(io::ErrorKind::InvalidData, "invalid checksum length"))?;
-    let expected_sum = u32::from_le_bytes(expected_sum);
+        let expected_sum: [u8; 4] = expected_sum
+            .try_into()
+            .map_err(|_| io::Error::new(io::ErrorKind::InvalidData, "invalid checksum length"))?;
+        let expected_sum = u32::from_le_bytes(expected_sum);
 
-    let output = match chunk_type {
-        ChunkType::Compressed => {
-            let uncompress_length = snap::raw::decompress_len(&data)?;
-            let mut out_buf = vec![0; uncompress_length];
-            let mut decoder = snap::raw::Decoder::new();
-            decoder.decompress(&data, &mut out_buf)?;
-            out_buf
+        self.out_buf.reset();
+        let mut out_buf = self.out_buf.get_mut();
+        out_buf.clear();
+        match chunk_type {
+            ChunkType::Compressed => {
+                let uncompress_length = snap::raw::decompress_len(data)?;
+                out_buf.resize(uncompress_length, 0);
+                let mut decoder = snap::raw::Decoder::new();
+                decoder.decompress(&data, &mut out_buf)?;
+            }
+            ChunkType::Uncompressed => out_buf.extend_from_slice(data),
+            _ => unreachable!(
+                "can only decode compressed or uncompressed chunks, not {:?}",
+                chunk_type
+            ),
+        };
+
+        let got_sum = crc32c::crc32c(&out_buf);
+        let got_sum = mask_crc(got_sum);
+        if expected_sum != got_sum {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidData,
+                "checksum mismatch",
+            ));
         }
-        ChunkType::Uncompressed => data,
-        _ => unreachable!(
-            "can only decode compressed or uncompressed chunks, not {:?}",
-            chunk_type
-        ),
-    };
 
-    let got_sum = crc32c::crc32c(&output);
-    let got_sum = mask_crc(got_sum);
-    if expected_sum != got_sum {
-        return Err(io::Error::new(
-            io::ErrorKind::InvalidData,
-            "checksum mismatch",
-        ));
+        Ok(())
     }
-
-    Ok(output)
 }
 
 #[derive(Debug)]
@@ -58,9 +72,8 @@ enum State {
     Buffering {
         remaining: usize,
         chunk_type: ChunkType,
-        buffer: Vec<u8>,
     },
-    Sending(PartialBuffer<Vec<u8>>),
+    Sending,
 }
 
 impl Default for State {
@@ -116,10 +129,11 @@ impl DecodeV2 for SnappyDecoder {
                         | ChunkType::ReservedSkippable(_)
                         | ChunkType::Padding => self.state = State::Skipping(data_frame_length),
                         ChunkType::Compressed | ChunkType::Uncompressed => {
+                            let in_buf = &mut self.in_buf;
+                            in_buf.clear();
                             self.state = State::Buffering {
                                 remaining: data_frame_length,
                                 chunk_type: header.chunk_type,
-                                buffer: Vec::with_capacity(data_frame_length),
                             }
                         }
                         ChunkType::ReservedUnskippable(chunk_type) => {
@@ -146,7 +160,6 @@ impl DecodeV2 for SnappyDecoder {
                 State::Buffering {
                     remaining,
                     chunk_type,
-                    buffer,
                 } => {
                     let input_buf = input.unwritten();
                     let boundary = (*remaining).min(input_buf.len());
@@ -154,7 +167,7 @@ impl DecodeV2 for SnappyDecoder {
 
                     *remaining -= input_buf.len();
 
-                    buffer.extend_from_slice(input_buf);
+                    self.in_buf.extend_from_slice(input_buf);
                     input.advance(input_buf.len());
 
                     if *remaining != 0 {
@@ -163,11 +176,11 @@ impl DecodeV2 for SnappyDecoder {
 
                     // We're done buffering, so let's decode the chunk
                     let chunk_type = *chunk_type;
-                    let buffer = mem::take(buffer);
-                    let output = decode_chunk(chunk_type, buffer)?;
-                    self.state = State::Sending(PartialBuffer::new(output))
+                    self.decode_chunk(chunk_type)?;
+                    self.state = State::Sending
                 }
-                State::Sending(buffer) => {
+                State::Sending => {
+                    let buffer = &mut self.out_buf;
                     output.copy_unwritten_from(buffer);
                     if buffer.unwritten().is_empty() {
                         self.state = State::ChunkHeader([0u8; 4].into())
@@ -181,7 +194,8 @@ impl DecodeV2 for SnappyDecoder {
 
     fn flush(&mut self, output: &mut WriteBuffer<'_>) -> std::io::Result<bool> {
         match &mut self.state {
-            State::Sending(buffer) => {
+            State::Sending => {
+                let buffer = &mut self.out_buf;
                 output.copy_unwritten_from(buffer);
                 if buffer.unwritten().is_empty() {
                     self.state = State::ChunkHeader([0u8; 4].into());
--- a/crates/compression-codecs/src/snappy/decoder.rs
+++ b/crates/compression-codecs/src/snappy/decoder.rs
@@ -90,7 +90,10 @@ impl Default for State {
 
 impl DecodeV2 for SnappyDecoder {
     fn reinit(&mut self) -> std::io::Result<()> {
-        *self = Self::new();
+        self.state = State::default();
+        self.in_buf.get_mut().clear();
+        self.in_buf.reset();
+        self.out_buf.reset();
         Ok(())
     }
 
@@ -108,7 +111,7 @@ impl DecodeV2 for SnappyDecoder {
                     }
 
                     let header = FrameHeader::parse(header.written())?;
-                    if let ChunkType::Stream = header.chunk_type {
+                    if matches!(header.chunk_type, ChunkType::Stream) {
                         self.state = State::Skipping(header.data_frame_length as usize)
                     } else {
                         return Err(std::io::Error::new(
--- a/crates/compression-codecs/src/snappy/decoder.rs
+++ b/crates/compression-codecs/src/snappy/decoder.rs
@@ -1,13 +1,12 @@
 use crate::snappy::{crc32c_masked, ChunkType, FrameHeader, MAX_BLOCK_SIZE, MAX_FRAME_SIZE};
 use crate::DecodeV2;
 use compression_core::util::{PartialBuffer, WriteBuffer};
-use std::convert::TryInto;
 use std::io;
 
 #[derive(Debug)]
 pub struct SnappyDecoder {
     state: State,
-    in_buf: Vec<u8>,
+    in_buf: PartialBuffer<Vec<u8>>,
     out_buf: PartialBuffer<Vec<u8>>,
 }
 
@@ -15,7 +14,7 @@ impl Default for SnappyDecoder {
     fn default() -> Self {
         Self {
             state: State::default(),
-            in_buf: Vec::with_capacity(MAX_FRAME_SIZE),
+            in_buf: PartialBuffer::new(Vec::with_capacity(MAX_FRAME_SIZE)),
             out_buf: PartialBuffer::new(Vec::with_capacity(MAX_BLOCK_SIZE)),
         }
     }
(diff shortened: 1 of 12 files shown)


NobodyXu:
> The "buffering" can be skipped if the input and/or output buffers are the right size. Should I implement that in this pull request?

I'm ok with either, though based on the size of the PR and that its CI is failing, I think getting a baseline implementation merged is a more sensible decision

Erouan50:
> > The "buffering" can be skipped if the input and/or output buffers are the right size. Should I implement that in this pull request?
> 
> I'm ok with either, though based on the size of the PR and that its CI is failing, I think getting a baseline implementation merged is a more sensible decision

That's fair. Let me fix the `proptest`. I forgot them...

crates/compression-codecs/src/snappy/decoder.rs
```diff
@@ -0,0 +1,203 @@
...
+    state: State,
+}
+
+impl SnappyDecoder {
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
+
+fn decode_chunk(chunk_type: ChunkType, mut buffer: Vec<u8>) -> std::io::Result<Vec<u8>> {
```
NobodyXu:
```suggestion
fn decode_chunk(chunk_type: ChunkType, buffer: &[u8]) -> std::io::Result<Vec<u8>> {
```

I think taking a mut reference is better, otherwise we lose the buffer on error, which could be used for debugging purposes

crates/compression-codecs/src/snappy/decoder.rs
```diff
@@ -0,0 +1,203 @@
...
+    let data = buffer.split_off(4);
```
NobodyXu: `Vec::split_off` creates a new vec containing all elements starting at index 4, which means it'd allocate and copy almost the same size as `buffer` to `data`

Using `<[u8]>::split_at` is much better

crates/compression-codecs/src/snappy/decoder.rs
```diff
@@ -0,0 +1,203 @@
...
+
+#[derive(Debug)]
+enum State {
+    StreamIdentifier(PartialBuffer<[u8; 4]>),
+    ChunkHeader(PartialBuffer<[u8; 4]>),
+    Skipping(usize),
+    Buffering {
+        remaining: usize,
+        chunk_type: ChunkType,
+        buffer: Vec<u8>,
```
NobodyXu: This buffer should probably be cached outside the `State` (and `reset` should clear it but not removing the heap allocation) to prevent frequent heap allocation.

crates/compression-codecs/src/snappy/decoder.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    let expected_sum: [u8; 4] = buffer
+        .try_into()
+        .map_err(|_| io::Error::new(io::ErrorKind::InvalidData, "invalid checksum length"))?;
+    let expected_sum = u32::from_le_bytes(expected_sum);
+
+    let output = match chunk_type {
+        ChunkType::Compressed => {
+            let uncompress_length = snap::raw::decompress_len(&data)?;
+            let mut out_buf = vec![0; uncompress_length];
```
NobodyXu: This should be cached like `buffer` as well (see below)

crates/compression-codecs/src/snappy/encoder.rs
```diff
@@ -0,0 +1,144 @@
+use crate::snappy::{crc32c_masked, ChunkType, FrameHeader, STREAM_FRAME};
+use crate::EncodeV2;
+use compression_core::util::{PartialBuffer, WriteBuffer};
+
+const MAX_BLOCK_SIZE: usize = 1 << 16;
+
+#[derive(Debug)]
+pub struct SnappyEncoder {
+    state: State,
+    chunk: Vec<u8>,
```
NobodyXu: Could wrap it in a `PartialBuffer`:

```suggestion
    chunk: PartialBuffer<Vec<u8>>,
```

crates/compression-codecs/src/snappy/encoder.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+#[derive(Debug)]
+enum State {
+    InitStream(PartialBuffer<&'static [u8]>),
+    Buffering,
+    Writing(PartialBuffer<Vec<u8>>),
+}
+
+fn compress_frame(buffer: &[u8]) -> std::io::Result<Vec<u8>> {
```
NobodyXu: Returning a `Vec` is so inefficient, please cache an `out_buf` like how `chunk` is cached

crates/compression-codecs/src/snappy/encoder.rs
```diff
@@ -0,0 +1,144 @@
...
+        data_frame_length: chunk_len as u64,
+    };
+
+    let mut frame = Vec::with_capacity(data.len() + 8);
+    let raw_header: [u8; 4] = header.into();
+    let raw_checksum: [u8; 4] = checksum.to_le_bytes();
+
+    frame.extend_from_slice(&raw_header);
+    frame.extend_from_slice(&raw_checksum);
+    frame.extend_from_slice(data);
```
NobodyXu: Mo absolutely not another Vec

You can extend the original vec with an empty slice of 8 bytes and then modify them

Erouan50:
> > The "buffering" can be skipped if the input and/or output buffers are the right size. Should I implement that in this pull request?
> 
> I'm ok with either, though based on the size of the PR and that its CI is failing, I think getting a baseline implementation merged is a more sensible decision

@NobodyXu I do agree to implement that in other PR. I fixed the tests except the one I mentioned in the description:

```
     Summary [  41.303s] 884 tests run: 882 passed, 2 failed, 0 skipped
        FAIL [   0.013s] async-compression::snappy snappy::futures::bufread::decompress::trailer
        FAIL [   0.008s] async-compression::snappy snappy::tokio::bufread::decompress::trailer
```

Do you have any opinion on how should I fix it? Or should I just skip this test for snappy?

crates/compression-codecs/src/snappy/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+
+impl From<u8> for ChunkType {
+    fn from(value: u8) -> Self {
+        match value {
+            0xFF => Self::Stream,
+            0x00 => Self::Compressed,
+            0x01 => Self::Uncompressed,
+            0xFE => Self::Padding,
+            0x02..=0x7f => Self::ReservedUnskippable(value),
+            0x80..=0xFD => Self::ReservedSkippable(value),
```
NobodyXu, line 44: Do we have to hardcode it here?

Ideally the `ChunkType` should be in `snap`

crates/compression-codecs/src/snappy/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+    slice[0] as u32 | (slice[1] as u32) << 8 | (slice[2] as u32) << 16
+}
+
+fn crc32c_masked(input: &[u8]) -> u32 {
+    let sum = crc32c::crc32c(input);
+    mask_crc(sum)
+}
+
+fn mask_crc(crc: u32) -> u32 {
+    (crc.wrapping_shr(15) | crc.wrapping_shl(17)).wrapping_add(0xA282EAD8)
```
NobodyXu, line 104: Another magic value, better in `snap`, I assume this is probably duplicate code?

crates/compression-codecs/src/snappy/decoder.rs
```diff
@@ -0,0 +1,203 @@
...
+        }
+        ChunkType::Uncompressed => data,
+        _ => unreachable!(
+            "can only decode compressed or uncompressed chunks, not {:?}",
+            chunk_type
+        ),
+    };
+
+    let got_sum = crc32c::crc32c(&output);
+    let got_sum = mask_crc(got_sum);
```
NobodyXu:
```suggestion
    let got_sum = crc32c_masked(&output);
```

NobodyXu (changes requested):
Thank you, I'm really unhappy with all the repeated vec creation which resulted in repeated copying and heap allocation

crates/compression-codecs/src/snappy/encoder.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    Ok(frame)
+}
+
+impl EncodeV2 for SnappyEncoder {
+    fn encode(
```
NobodyXu, line 102: Why do we have to cache the input into a 32KB chunk? This won't work well in network and would force a flush

Won't it be easier to just act like other encoder here, store the `snap::raw::Encoder`, for whatever input, just compress it and write to the `output`?

Yes we might have to cap the `input` at 32K, and might need to further cap the `input` if `output` is too small (use `compress_len` to guess the compression ratio and then cap `input`), but won't it be simpler than adding extra buffer?

NobodyXu (changes requested):
Question regarding why encoder is designed with extra buffering

I can understand that the decoder needs it as we cannot control on bow the chunks are created

NobodyXu:
> Do you have any opinion on how should I fix it? Or should I just skip this test for snappy?

Does snappy have a "file header" that recorded the size of the entire stream?

If we have it, then we can configure the decoder to stop decoding once that limit is matched, and only way to continue would be `reinit`

Otherwise we cannot fix it

crates/compression-codecs/src/snappy/encoder.rs
Erouan50: That a good question. I followed the [Write][1] implementation from the crate snap. And I think other snappy libs (Python, Go, etc...) do the same by trying to accumulate as much as they can before compressing. 

I didn't know we could use `compress_len` to cap the input to not overfill the output. I'll take a look. However, I feel the compression is better if we compress as much as we can in one call.

[1]: https://github.com/BurntSushi/rust-snappy/blob/master/src/write.rs

Erouan50:
> > Do you have any opinion on how should I fix it? Or should I just skip this test for snappy?
> 
> Does snappy have a "file header" that recorded the size of the entire stream?
> 
> If we have it, then we can configure the decoder to stop decoding once that limit is matched, and only way to continue would be `reinit`
> 
> Otherwise we cannot fix it

No, the first "stream" chunk is just there to mark the start of the file, but it doesn't include the length of the whole stream.

Erouan50:
> Thank you, I'm really unhappy with all the repeated vec creation which resulted in repeated copying and heap allocation

You're right, I apologize for that... I tried to apply your suggestions. It's not perfect (I don't like the `reset()` + `clear()` + `resize()` in the decoder for example), but I hope I'm going to the right direction?

crates/compression-codecs/src/snappy/mod.rs
Erouan50: It is but the enum is not public.

crates/compression-codecs/src/snappy/mod.rs
Erouan50: Yes it's duplicate code. The function is not public.

NobodyXu:
> No, the first "stream" chunk is just there to mark the start of the file, but it doesn't include the length of the whole stream.

Do we have a trailer/footer that marks the end of the stream?

If no, then we have to remove the test for snappy

crates/compression-codecs/src/snappy/encoder.rs
NobodyXu: In that case I'm fine with the current implementation.

crates/compression-codecs/src/snappy/mod.rs
NobodyXu: I think it'd be better if `snap` makes it public under the `raw` module

crates/compression-codecs/src/snappy/mod.rs
NobodyXu: Can we have it public under `snap::raw` module?

Hardcoding it here really feels wrong

crates/compression-codecs/src/snappy/decoder.rs
```diff
@@ -0,0 +1,216 @@
...
+        let out_buf = self.out_buf.get_mut();
+        out_buf.clear();
+        match chunk_type {
+            ChunkType::Compressed => {
+                let uncompress_length = snap::raw::decompress_len(data)?;
+                out_buf.resize(uncompress_length, 0);
+                let mut decoder = snap::raw::Decoder::new();
+                decoder.decompress(data, out_buf)?;
+            }
+            ChunkType::Uncompressed => out_buf.extend_from_slice(data),
```
NobodyXu: We should skip this copy, by putting the chunk type into  a new variant `UncompressedCopy` so that it'd copy from `data`, along with an offset, initially set to 4 to skip the header.

Using that offset we can construct a `PartialBuffer`

crates/compression-codecs/src/snappy/decoder.rs
```diff
@@ -0,0 +1,216 @@
...
+
+        self.out_buf.reset();
+        let out_buf = self.out_buf.get_mut();
+        out_buf.clear();
+        match chunk_type {
+            ChunkType::Compressed => {
+                let uncompress_length = snap::raw::decompress_len(data)?;
+                out_buf.resize(uncompress_length, 0);
+                let mut decoder = snap::raw::Decoder::new();
+                decoder.decompress(data, out_buf)?;
```
NobodyXu, line 46: It's a shame that `snap` does not support writing to uninitialised buffer, zeroing the whole vec would indeed take quite some time

crates/compression-codecs/src/snappy/decoder.rs
```diff
@@ -0,0 +1,216 @@
...
+
+impl Default for State {
+    fn default() -> Self {
+        State::StreamIdentifier(PartialBuffer::new([0; 4]))
+    }
+}
+
+impl DecodeV2 for SnappyDecoder {
+    fn reinit(&mut self) -> std::io::Result<()> {
+        *self = Self::new();
```
NobodyXu:
```suggestion
        self.state = State::default();
        self.in_buf.clear();
        self.out_buf.reset();
```

crates/compression-codecs/src/snappy/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+
+        let mut header = [0u8; 4];
+        header[0] = u8::from(value.chunk_type);
+        // We're writing a little endian u24 from an u32 by removing the latest significant byte
+        header[1..4].copy_from_slice(&frame_length.to_le_bytes()[..3]);
+        header
+    }
+}
+
+pub fn read_u24_le(slice: &[u8; 3]) -> u32 {
```
NobodyXu:
```suggestion
fn read_u24_le(slice: &[u8; 3]) -> u32 {
```

This function is used by snappy submodules and doesn't have to be public

crates/compression-codecs/src/snappy/mod.rs
```diff
@@ -0,0 +1,104 @@
...
+
+impl FrameHeader {
+    fn parse(input: &[u8]) -> io::Result<Self> {
+        let (header_part, _): (&[u8; 4], _) = input.split_first_chunk().ok_or(io::Error::new(
+            io::ErrorKind::UnexpectedEof,
+            "Need a least 4 bytes to parse the frame's header",
+        ))?;
+
+        let chunk_type = ChunkType::from(header_part[0]);
+        // SAFETY: header_part is guaranteed to have at least 4 bytes due to split_first_chunk
```
NobodyXu:
```suggestion
        // header_part is guaranteed to have at least 4 bytes due to split_first_chunk
```

Safety is for `unsafe` block and function

crates/compression-codecs/src/snappy/decoder.rs
```diff
@@ -0,0 +1,216 @@
...
+    }
+}
+
+#[derive(Debug)]
+enum State {
+    StreamIdentifier(PartialBuffer<[u8; 4]>),
+    ChunkHeader(PartialBuffer<[u8; 4]>),
+    Skipping(usize),
+    Buffering {
+        remaining: usize,
```
NobodyXu:
```suggestion
    Skipping(NonZeroUize),
    Buffering {
        remaining: NonZeroUsize,
```

Using `NonZeroUsize` would make it clear the invariant of the state

crates/compression-codecs/src/snappy/encoder.rs
```diff
@@ -0,0 +1,158 @@
...
+        let checksum = crc32c_masked(in_buffer);
+
+        self.out_buf.reset();
+        let out_buf = self.out_buf.get_mut();
+        out_buf.clear();
+        let max_compress_size = snap::raw::max_compress_len(in_buffer.len());
+        out_buf.resize(max_compress_size + 8, 0);
+
+        let mut encoder = snap::raw::Encoder::new();
+        let compress_data = encoder.compress(in_buffer, &mut out_buf[8..])?;
```
NobodyXu, line 40: It's a shame `snap::raw` doesn't support uninitialised buffer

crates/compression-codecs/src/snappy/encoder.rs
```diff
@@ -0,0 +1,158 @@
...
+
+        let chunk_type = if compress_data >= in_buffer.len() - (in_buffer.len() / 8) {
+            out_buf.clear();
+            out_buf.resize(in_buffer.len() + 8, 0);
+            out_buf[8..].copy_from_slice(in_buffer);
```
NobodyXu: We should have a new state `UncompressedCopy` instead of doing this intermediate copy

NobodyXu (changes requested):
Thanks!

My feedback is that the copying around of uncompressed sections is unnecessary and should be directly copied into the `output`

And that `snap::raw` should have more public items to prevent duplicate, and that it needs to support uninitialised buffer for maximum efficiency

NobodyXu:
> It's not perfect (I don't like the `reset()` + `clear()` + `resize()` in the decoder for example), but I hope I'm going to the right direction?

Yeah it's the right direction, it looks ugly because `snap::raw` does not support uninitialised buffer, speaking of which the `{de, en}code_vec` is also very inefficient due to that

crates/compression-codecs/src/snappy/decoder.rs
Erouan50: I don't think I can do that because remaining and skipping can be 0.

crates/compression-codecs/src/snappy/mod.rs
Erouan50: I agree. I'll open a pull request to the repository, but I can't guarantee they'll accept the changes.

crates/compression-codecs/src/snappy/mod.rs
Erouan50: I'll open a pull request.

crates/compression-codecs/src/snappy/decoder.rs
Erouan50: I'll try to support it. It could be a good challenge to tackle :).

crates/compression-codecs/src/snappy/encoder.rs
Erouan50: I'll try to support it as well.

Erouan50:
> > It's not perfect (I don't like the `reset()` + `clear()` + `resize()` in the decoder for example), but I hope I'm going to the right direction?
> 
> Yeah it's the right direction, it looks ugly because `snap::raw` does not support uninitialised buffer, speaking of which the `{de, en}code_vec` is also very inefficient due to that

I agree. And even the whole decoding/encoding of the frames should be in the snap crate. I did it here because the crate supports only io::Read and io::Write.

If you want, I can give a shot at contributing to the crate with all the changes you mentioned and we pause/close that pull request?

NobodyXu:
> If you want, I can give a shot at contributing to the crate with all the changes you mentioned and we pause/close that pull request?

Yeah it'd be awesome great to get them upstreamed, thank you!

If you can share the PR/tracking issue so that I can be in the loop that'd even better

crates/compression-codecs/src/snappy/decoder.rs
NobodyXu: That's the neat part: if it's 0, then we just skip that stage and go to the next stage

crates/compression-codecs/src/snappy/mod.rs
NobodyXu: It's ok! We should at least try it and I believe the motivation for exposing this is reasonable

crates/compression-codecs/src/snappy/encoder.rs
```diff
@@ -0,0 +1,180 @@
...
+use crate::EncodeV2;
+use compression_core::util::{PartialBuffer, WriteBuffer};
+
+const MAX_BLOCK_SIZE: usize = 1 << 16;
+
+#[derive(Debug)]
+pub struct SnappyEncoder {
+    state: State,
+    in_buf: PartialBuffer<Vec<u8>>,
+    out_buf: PartialBuffer<Vec<u8>>,
```
NobodyXu, line 11:
```suggestion
    out_buf: Vec<u8>,
```

I think `out_buf` doesn't have to beba `PartialBuffer` as we don't use that functionality

crates/compression-codecs/src/snappy/decoder.rs
```diff
@@ -0,0 +1,245 @@
+use crate::snappy::{crc32c_masked, ChunkType, FrameHeader, MAX_BLOCK_SIZE, MAX_FRAME_SIZE};
+use crate::DecodeV2;
+use compression_core::util::{PartialBuffer, WriteBuffer};
+use std::io;
+
+#[derive(Debug)]
+pub struct SnappyDecoder {
+    state: State,
+    in_buf: PartialBuffer<Vec<u8>>,
+    out_buf: PartialBuffer<Vec<u8>>,
```
NobodyXu, line 13:
```suggestion
    out_buf: Vec<u8>,
```

I don't think `PartialBuffer` functionality is used here

NobodyXu (changes requested):
Thanks just some minor nit

crates/compression-codecs/src/snappy/encoder.rs
Erouan50: I use the PartialBuffer internal index to keep track of the copy [here][1] through [here][2], but I can do it another way if you prefer.

[1]: https://github.com/Nullus157/async-compression/pull/466/changes#diff-bef7bd55364943f21f8ee3aa05fd0f3ad93e92f000a83d21f41a0fed23602736R85
[2]: https://github.com/Nullus157/async-compression/pull/466/changes#diff-bef7bd55364943f21f8ee3aa05fd0f3ad93e92f000a83d21f41a0fed23602736R143

crates/compression-codecs/src/snappy/decoder.rs
Erouan50: I use the PartialBuffer [here][1] to keep track of the copy, we may have another way to keep track of the copy?

[1]: https://github.com/Nullus157/async-compression/pull/466/changes#diff-61ceda7b8462145902249d837575daaafb12aa7d9dc44915b37c9577e6862d7cR203

NobodyXu: Sorry I missed it

crates/compression-codecs/src/snappy/encoder.rs
NobodyXu: Sorry I missed it

NobodyXu (commented):
I think it looks good to me now, except for the `NonZeroUsize` and the upstream changes required (more `pub` and uninitialised buffer

cc @robjtede

crates/compression-codecs/src/snappy/decoder.rs
Erouan50: Indeed! That's better that way.

Erouan50:
> I think it looks good to me now, except for the `NonZeroUsize` and the upstream changes required (more `pub` and uninitialised buffer
> 
> cc @robjtede

Here is my pull request to support uninitialized buffer: https://github.com/BurntSushi/rust-snappy/pull/65

NobodyXu:
Looks like we've identified some nightly rustc bugs?

PR #467, opened 2026-04-21
Include license files in published crates

ankane:
Hi, this PR symlinks the license files so they're included in published crates (which is nice for tools like [`cargo vendor`](https://doc.rust-lang.org/cargo/commands/cargo-vendor.html)).


ankane:
Thanks @robjtede

PR #470, opened 2026-07-15
Fix hang when decoding a corrupt subsequent zstd frame

zanieb:
Fix an infinite loop when decoding concatenated zstd frames with `multiple_members(true)` and a later frame is corrupt.

After completing a frame, the bufread decoder incorrectly treats the next decode as an empty-input poll even when input remains buffered, suppressing the corruption error and allowing the decoder to spin without making progress. Only request another poll when the current buffer is exhausted, ensuring errors from subsequent frames are propagated.


NobodyXu (approved):
Thank you, LGTM! Will cut a release after merging

PR #477, opened 2026-09-01
Enforce WriteBuffer bounds checks in release builds

### LINKED ISSUE #474: Safe uninitialized memory access via `WriteBuffer::advance`
I'm reporting this publicly because, while it is memory unsafety, I don't believe it's exploitable in practice. In particular, I couldn't find any way for an attacker-controlled compressed stream to manipulate the `advance()` API.

MRE:

```rust
use compression_core::util::WriteBuffer;

fn main() {
    // Heap-backed spare capacity remains reliably poisoned under MemorySanitizer.
    let mut allocation = Vec::<u8>::with_capacity(8);
    let storage = &mut allocation.spare_capacity_mut()[..8];
    let mut output = WriteBuffer::new_uninitialized(storage);

    // This safe call succeeds in release mode because the initialization check
    // is only a debug_assert!.
    output.advance(1);

    // The index is in bounds, but the referenced byte is uninitialized.
    let byte = output.written()[0];
    println!("uninitialized byte: {byte}");
}
```

When built in the release profile, this advances into uninitialized memory, since the underlying `advance` API only uses `debug_assert!` for its bounds checks:

https://github.com/Nullus157/async-compression/blob/3a711b04f31cad68cccf0d09d3115b38556fc0da/crates/compression-core/src/util.rs#L154-L162

Similarly, the `written()` API only uses a debug-build guard when exposing the initialized slice:

https://github.com/Nullus157/async-compression/blob/3a711b04f31cad68cccf0d09d3115b38556fc0da/crates/compression-core/src/util.rs#L125-L128

The other APIs that assume initialization are correctly marked as `unsafe`, e.g. `assume_init_and_advance` has the same `debug_assert` but is marked as `unsafe` to emphasize that it's the caller's responsibility to maintain the `WriteBuffer`'s internal invariants.

I think there are two possible solutions here:

- Make `advance()` and `written()` into `unsafe` APIs, making it clear that the caller is required to enforce bounds checks.
- Replace `debug_assert!` with `assert!` in these safe APIs, so that the bounds check happens in release builds as well.

I _think_ that second solution is probably for the best, since it doesn't change the public APIs in any way. I suspect the performance risks would also be negligible, but I haven't measured them.

woodruffw:
Previously the `advance` and `written` APIs used `debug_assert` for bounds checks, which wouldn't be enforced in release builds. They now use `assert`, and additionally the `advance` asserts have been rewritten to avoid an overflow on addition.

I've also added some tests for the panicking and non-panicking cases with these APIs.

Fixes #474.

### DIFF
--- a/crates/compression-core/CHANGELOG.md
+++ b/crates/compression-core/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Enforce `WriteBuffer::advance` and `WriteBuffer::written` initialization checks in release builds to prevent safe access to uninitialized memory ([#474](https://github.com/Nullus157/async-compression/issues/474)).
+
 ## [0.4.32](https://github.com/Nullus157/async-compression/compare/compression-core-v0.4.31...compression-core-v0.4.32) - 2026-04-21
 
 ### Other
--- a/crates/compression-core/CHANGELOG.md
+++ b/crates/compression-core/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


NobodyXu (approved):
Thank you!

PR #479, opened 2026-09-02
Fix safety contract on assume_init, use asserts

### LINKED ISSUE #475: Doc discrepancy/uninitialized access via `advance`
This one is pretty similar to #474, although slightly different in root cause. Like #474 I don't think there's an immediate route to exploitation (AFAICT none of the compression implementations cal `assume_init` multiple times), so I'm opening it publicly.

MRE:

```rust
use compression_core::util::WriteBuffer;

fn main() {
    // Heap-backed spare capacity remains reliably poisoned under MemorySanitizer.
    let mut allocation = Vec::<u8>::with_capacity(6);
    let storage = &mut allocation.spare_capacity_mut()[..6];
    for byte in &mut storage[..4] {
        byte.write(0x41);
    }

    let mut output = WriteBuffer::new_uninitialized(storage);

    // SAFETY: Bytes zero through three really are initialized.
    unsafe { output.assume_init(4) };

    // SAFETY: written_len is still zero, so the first two unfilled bytes are
    // bytes zero and one. They are initialized. The docs say this overlapping
    // call should do nothing.
    unsafe { output.assume_init(2) };

    // The implementation added both calls, so this safe advance now succeeds
    // even though bytes four and five remain uninitialized.
    output.advance(6);
    let byte = output.written()[4];
    println!("uninitialized byte: {byte}");
}
```

The docs for `assume_init` say that `n` only affects the _first_ `n` bytes in the buffer, i.e. is not additive across calls. However, the implementation is additive:

https://github.com/Nullus157/async-compression/blob/3a711b04f31cad68cccf0d09d3115b38556fc0da/crates/compression-core/src/util.rs#L181-L194

As a result, an API consumer who obeys the documented safety contract can accidentally access uninitialized memory.

Two possible solutions:

- Update the documented safety contract to emphasize that `assume_init(n)` assumes that the _next_ `n` bytes are initialized, beyond any bytes already marked initialized via prior calls.
- Update `assume_init` so that it sets `self.initialized = max(self.initialized, index + n)`

(These two solutions are mutually exclusive, since the first updates the safety contract and the latter fixes the implementation to match the current contract.)

woodruffw:
This follows #477, and closes #475.

TL;DR: `assume_init`'s safety contract said that calls were not additive, but the implementation was additive. This fixes the implementation and clarifies the safety contract, plus switches `debug_assert` to `assert` for the invariant check.

I've also added similar safety/panic contracts and switched to `assert` in the other APIs in the module.

### DIFF
--- a/crates/compression-core/CHANGELOG.md
+++ b/crates/compression-core/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix `WriteBuffer::assume_init` to track initialization relative to the written position without adding overlapping calls, and enforce bounds checks for initialization setters in release builds ([#475](https://github.com/Nullus157/async-compression/issues/475)).
 - Enforce `WriteBuffer::advance` and `WriteBuffer::written` initialization checks in release builds to prevent safe access to uninitialized memory ([#477](https://github.com/Nullus157/async-compression/pull/477)).
 
 ## [0.4.32](https://github.com/Nullus157/async-compression/compare/compression-core-v0.4.31...compression-core-v0.4.32) - 2026-04-21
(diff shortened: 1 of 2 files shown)


NobodyXu (approved):
Thank you! Will cut a release immediately

PR #483, opened 2026-09-08
Return UnexpectedEof when deflate64 does not progress

woodruffw:
This is technically a denial-of-service vector, but I'm opening it publicly because I think it's _very_ niche (Deflate64 is itself niche).

Deflate64 uses a NUL byte to indicate the start of a stored (i.e. non-compressed) block. However, if no data actually follows that marker, the implementation (on the async adapter side) would previously spin forever since `finish` would only signal "not finished" rather than a premature EOF.

The fix is to have the decompression layer check for actual read progress, and then return UnexpectedEof rather than a finished/not-finished boolean state when no more progress can be made.

### DIFF
--- a/crates/async-compression/Cargo.toml
+++ b/crates/async-compression/Cargo.toml
@@ -100,6 +100,10 @@ name = "deflate"
 required-features = ["deflate"]
 
 [[test]]
+name = "deflate64"
+required-features = ["deflate64", "tokio"]
+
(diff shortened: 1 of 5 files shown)


NobodyXu (approved):
Thank you! Will cut a new release immediately

woodruffw:
Thanks @NobodyXu!

Issue #485, opened 2026-09-10
Zstd decoding fails after repeated Poll::Pending with compression-codecs ≥ 0.4.40


bvinc:
Valid Zstd input fails to decode when the input stream returns Poll::Pending repeatedly between chunks:

```
Operation made no progress over multiple calls, due to input being empty
```

Reproduced with async-compression 0.4.46 / compression-codecs 0.4.41.

```toml
[package]
name = "zstd-pending-repro"
version = "0.1.0"
edition = "2021"

[dependencies]
async-compression = { version = "=0.4.46", features = ["tokio", "zstd"] }
compression-codecs = { version = "=0.4.41", features = ["zstd"] }
tokio = { version = "=1.53.1", features = ["rt", "macros", "io-util"] }
tokio-util = { version = "=0.7.19", features = ["io"] }
futures = "0.3"
```

src/main.rs
```rust
use std::{io, task::Poll};

use async_compression::tokio::bufread::{ZstdDecoder, ZstdEncoder};
use futures::stream;
use tokio::io::AsyncReadExt;
use tokio_util::io::StreamReader;

#[tokio::main(flavor = "current_thread")]
async fn main() -> io::Result<()> {
    // Use 0 as a passing control; 32 deterministically triggers the bug.
    let mut pending = std::env::args()
        .nth(1)
        .map(|value| value.parse::<usize>())
        .transpose()
        .map_err(|error| io::Error::new(io::ErrorKind::InvalidInput, error))?
        .unwrap_or(32);
    println!("Pending polls between compressed chunks: {pending}");

    let data: Vec<u8> = (0u32..4096)
        .flat_map(|i| i.wrapping_mul(2654435761).to_le_bytes())
        .collect();

    let mut encoded = Vec::new();
    ZstdEncoder::new(data.as_slice())
        .read_to_end(&mut encoded)
        .await?;
    assert!(encoded.len() > 1024);

    let mut first = true;
    let mut done = false;
    let source = stream::poll_fn(move |cx| {
        if first {
            first = false;
            return Poll::Ready(Some(Ok::<_, io::Error>(io::Cursor::new(
                encoded[..1024].to_vec(),
            ))));
        }
        if pending > 0 {
            pending -= 1;
            // Model repeated legal polling while waiting for the next chunk.
            cx.waker().wake_by_ref();
            return Poll::Pending;
        }
        if !done {
            done = true;
[...]

PR #486, opened 2026-09-10
fix(zstd): avoid no-progress errors while awaiting input

bvinc:
Proposed fix for issue #485 

Skip repeated empty-input decoding after buffered output is drained. Add Tokio and futures-io regression tests.


crates/compression-codecs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Avoid poisoning the Zstd decoder when repeatedly polled while waiting for more input.
```
NobodyXu, line 12: Please include PR number in the changelog

Btw this changelog is auto generated, so you can skip this if you don't want any special changelog message

NobodyXu (changes requested):
Thank you, just want the changelog formatted

crates/compression-codecs/CHANGELOG.md
bvinc: I removed the CHANGELOG change

NobodyXu (approved):
Thanks, will cut a release after merging this

